PR #21, opened 2022-09-09
Solution

MaxManis:


src/bullsAndCows.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable */
+'use strict';
+/* eslint-enable */
+
+// - The bullsAndCows module is used to compare
+// |-- the user's and computes's strings
+// |-- and count the Bulls\Cows result:
+function bullsAndCows(user, cp) {
```
Denchepornyuk, line 8: `cp` isn't a best name

src/bullsAndCows.js
```diff
@@ -0,0 +1,23 @@
...
+  let bulls = 0;
+  let cows = 0;
+
+  for (let i = 0; i < cp.length; i++) {
+    if (cp[i] === user[i]) {
+      bulls++;
+    } else if (cp.split('').includes(user[i])) {
```
Denchepornyuk, line 15: do you need split here?
includes also work with strings

Denchepornyuk (approved):

PR #28, opened 2022-09-26
solution

Dina-Samoilova:


src/setRandomNumber.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+let number = Math.round(Math.random() * 10);
```
qoa11a, line 3: It's possible that this generates a number 10.

qoa11a (changes requested):

PR #35, opened 2022-10-12
Solution

bogdana18:


src/bull.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+// eslint-disable-next-line no-console
```
loralevitska, line 3: there is no console in this file. Remove comment

src/bull.js
```diff
@@ -0,0 +1,20 @@
...
+const { randomNumber } = require('./getRandom');
+
+function bull(param) {
+  let count = 0;
+  const arrUser = param.split('');
+  const arrPC = randomNumber.split('');
+
+  arrUser.filter((item, index) => {
```
loralevitska, line 11: What exactly are you filtering if you return `undefined` for each element of the array. Use `forEach` or loops

loralevitska (changes requested):

PR #37, opened 2022-10-13
solution

tiserett:
[LINK](https://codesandbox.io/s/github/tiserett/node_bulls-and-cows/tree/develop?file=/src/app.js) You can play here ;)
Fork it, then open terminal and write 'node src/app.js'


src/app.js
```diff
@@ -1 +1,35 @@
...
+const { bull } = require('./bull');
+const { cow } = require('./cow');
+const { randomNumber } = require('./getRandomNumber');
+
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+console.log(randomNumber);
```
sergii-nosachenko, line 14:
```suggestion
```
It's a secret)

src/cow.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const { randomNumber } = require('./getRandomNumber');
+
+function cow(userInput) {
+  let count = 0;
+
+  const arrInput = userInput.split('');
+  const arrNumber = randomNumber.split('');
```
sergii-nosachenko, line 9:
```suggestion
```
Useless code. randomNumber has length and includes() method. Also you can get chars by indexes. So you don;t need arrays at all.

src/bull.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const { randomNumber } = require('./getRandomNumber');
+
+function bull(userInput) {
+  let count = 0;
+
+  const arrInput = userInput.split('');
+  const arrNumber = randomNumber.split('');
```
sergii-nosachenko, line 9:
```suggestion
```
Useless code. randomNumber has length and includes() method. Also you can get chars by indexes. So you don;t need arrays at all.

src/bull.js
```diff
@@ -0,0 +1,20 @@
...
+
+  for (let i = 0; i < arrNumber.length; ++i) {
```
sergii-nosachenko, line 11:
```suggestion
  for (let i = 0; i < arrNumber.length; i++) {
```
It's safer to use postfix increment. It always returns incremented value, prefix increments laso but returns previous value. Please, replace it all over your code.

src/getRandomNumber.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function getRandomNumber() {
+  let result = '';
+
+  for (let i = 0; result.length !== 4; i++) {
```
sergii-nosachenko, line 6: Genius loop condition, I like it!

src/app.js
```diff
@@ -1 +1,35 @@
...
+
+      question();
+    } else if (userInput === randomNumber) {
+      console.log(`You won!`);
+      terminal.close();
+    } else {
+      const bulls = bull(userInput);
+      const cows = cow(userInput);
+
+      console.log(`The result is ${bulls} bull('s) and ${cows} cow('s)`);
```
sergii-nosachenko, line 29: Optional, but will be good for app user:
1) `'s` can be added for numbers > 1 dynamically by condition;
2) don't show 0's
3) if no cows and no bulls - show another message

sergii-nosachenko (changes requested):
Good job, I like it)
But there're some minor issues you need to fix.

src/app.js
tiserett: no way... I forgot to delete it 🤫🤫🤫

src/bull.js
tiserett: By the way, about the prefix increment, I can say that it is safe to use this thing in loops, since this operation is performed at the end of the iteration, and with a prefix increment, a copy of the variable is not created to add a value to it, how it is done with postfix

sergii-nosachenko (approved):
Awesome 🔥

src/app.js
```diff
@@ -1 +1,33 @@
+/* eslint-disable no-console */
```
AlinaMatuschak, line 1: don't disable eslint

AlinaMatuschak (approved):

PR #38, opened 2022-10-13
Solution

Jamal-Amoian:


src/app.js
```diff
@@ -1 +1,53 @@
...
+    if (isExist) {
+      str += 'cows ';
+
+      continue;
+    }
+  }
+
+  console.log('Random numbers: ' + randomNumbers);
+
+  console.log(str);
```
AlinaMatuschak, line 50: will be nice on the console to show numbers of bull and cows

AlinaMatuschak (approved):
Good, but you need to return the string `The result is one bull and 2 cows`. So need to create separate modules for calculating bulls and cows

PR #42, opened 2022-10-14
Solution

VolodymyrChuyko:


NataliyaAn (approved):
Good decision!
If you add the opportunity to add "0", it will be better.
Now User sees the error 
```
Enter your number to play or "q" to exit: 2904
Your should enter a number of 4different digits!
```

AndriiRut (approved):
Good job! 👍🏻

PR #48, opened 2022-10-25
Develop

Dima-Semenov:


src/app.js
```diff
@@ -1 +1,57 @@
...
+      } else {
+        console.log(
+          `You guessed ${countOfBulls} bulls and ${countOfCows} cows\n`,
+        );
+        countOfAttempts += 1;
+        typeOfQuestions = KEY_QUESTIONS.try_again;
+        startGame();
+      }
+    } else {
+      console.log('The number of characters must be 4\n');
```
ArturErmolenko, line 44: Do not rely on the unknown string, make constants(enum) for this

src/app.js
```diff
@@ -1 +1,57 @@
...
+      startGame();
+    }
+  });
+
+console.log('\nRules:');
+console.log('You need to guess the 4-digit number that the computer guessed');
+console.log(`'bull' - guessed digit is on its place (index)`);
```
ArturErmolenko, line 51: same here

src/countBullsAndCows.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+  }
+
+  return {
+    isCorrect: countOfBulls === 4,
+    countOfBulls,
+    countOfCows,
+  };
+};
+
```
ArturErmolenko, line 47: interesting solution 👍 
I think it would be possible to do a cycle for and use includes ,this would reduce the amount of code, but not critical JFYI

ArturErmolenko (approved):
Good job 👍

PR #50, opened 2022-11-02
looks good for me

ZUBOV-ILLIA:


src/generateNumbers.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+function getRandomNumber() {
+  const randomNumber = Math.random().toString();
+  let result = '';
+
+  for (let i = randomNumber.length - 1; i >= 0; i--) {
```
alexandra-protyanova, line 7: I would use`while` loop here 😉

PR #53, opened 2022-11-16
bulls and cows implemented

artem-musii:


src/generateNumber.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const generateNumber = () => {
+  const randoms = [];
+
+  while (randoms.length < 4) {
+    const randomDigit = Math.floor(Math.random() * 10 + 1);
```
anastasiia-nahaiska, line 7: I think now you generate numbers from 1 to 10, because we use this formula 
'Math.floor(Math.random() * (max - min + 1) + min)'
So if you want to generate numbers from 1 to 9 (we must generate digits) the formula will be looks as
'Math.floor(Math.random() * 9 + 1)'
but maybe I'm wrong)

src/interactWithTerminal.js
```diff
@@ -0,0 +1,33 @@
...
+  output: process.stdout,
+});
+
+const interactWithTerminal = () => {
+  terminal.question('Try to guess: ', (guess) => {
+    for (const char of guess) {
+      if ((guess.indexOf(char) !== guess.lastIndexOf(char))
+      || isNaN(+char)
+      || guess.length !== 4) {
+        throw new Error('Incorrect input');
```
anastasiia-nahaiska, line 19: Would be better to add some text about why it is incorrect input or which input will be correct

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,40 @@
...
+
+      if (numbers[+randomNumber[i]] < 0) {
+        cows++;
+      }
+
+      numbers[+randomNumber[i]]++;
+      numbers[+guess[i]]--;
+    }
+  }
+
```
anastasiia-nahaiska, line 28: I think this function may looks easier because we always generate different digits) So if it isn't a bull, we need only check if a secret includes guess[i] to find out if it a cow or no))

anastasiia-nahaiska (approved):
GJ!) I left a few comments)

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,33 @@
...
+  let bulls = 0;
+
+  for (let i = 0; i < guess.length; i++) {
+    if (randomNumber[i] === guess[i]) {
+      bulls++;
+    } else {
+      if (randomNumber.includes(guess[i])) {
+        cows++;
+      }
+    }
```
silchyn, line 19:
```suggestion
    } else  if (randomNumber.includes(guess[i])) {
      cows++;
    }
```

silchyn (approved):
Cool!

PR #61, opened 2022-11-17
Develop

TheVovchik:
Solution on task bulls and cows


src/generateSecretNumber.js
```diff
@@ -0,0 +1,18 @@
+// You may think that I invented a bicycle
```
silchyn, line 1: Isn't it easier to generate one random number between 1000 and 9999?

silchyn (changes requested):

silchyn (approved):
Cool!

PR #66, opened 2022-11-21
added (calculate, terminal, randomGenerator), implemented solution

SerhiiRomaniukWD:


src/generateRandomNumber.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const generateRandomNumber = (min, max) => {
+  return Math.floor(Math.random() * (max - min) + min);
+};
```
Andrew-Skoryk, line 5: Capture
the program should randomly generate a number of 4 _different_ digits

Andrew-Skoryk (changes requested):
GJ! Needed a few fixes: random number should not be totally random=) Also, I would change the name of 'buildDialogue.js' to ' terminal' or 'gameIterface'. Moreover, you can move game logic to that file, and app.js would only start the game, but it's up to you.

PR #78, opened 2023-01-25
Release of a game

tarasavage:


src/app.js
```diff
@@ -1 +1,31 @@
...
+      || userInput.split('').some((e, i, arr) => arr.indexOf(e) !== i))) {
+      console.log('Wrong input!');
+      numberIsCorrect = false;
+      game(randomInput);
+    }
+
+    const animals = compareNumbers(randomInput, userInput);
+
+    if (animals.bulls === 4) {
+      console.log('Congratulations! You won!');
```
polosanya, line 24: Let's try not to use console log. Just return the string with proper message instead.
Fix all cases

polosanya (changes requested):
Let's try to pass the tests)

src/app.js
```diff
@@ -1 +1,34 @@
...
+
+const random = generateNumber();
+
+(function game(randomInput) {
+  terminal.question('Please, enter the number: ', (userInput) => {
+    let numberIsCorrect = true;
+
+    if ((userInput.length !== 4
+      || userInput.split('').some(num => isNaN(+num))
+      || userInput.split('').some((e, i, arr) => arr.indexOf(e) !== i))) {
```
Bohdan-Maksymiuk, line 15: lets split these conditions into separate variables and then use here

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,23 @@
...
+
+  const digitsRandom = `${random}`.split('');
+  const digitsInput = `${input}`.split('');
+
+  for (let i = 0; i < digitsRandom.length; i++) {
+    if (digitsInput[i] === digitsRandom[i]) {
+      count.bulls++;
+    } else if (digitsRandom.includes(digitsInput[i])) {
+      count.cows++;
+    }
```
Bohdan-Maksymiuk, line 17: avoid such nesting `else if` conditions
This code is hard to read and maintain

src/app.js
```diff
@@ -1 +1,34 @@
...
+
+    if (animals.bulls === 4) {
+      terminal.close();
+
+      return 'Congratulations! You won!';
+    } else if (numberIsCorrect) {
+      terminal.close();
+
+      return `Bulls: ${animals.bulls} Cows: ${animals.cows}`;
+    }
```
Bohdan-Maksymiuk, line 32: avoid such nesting `else if` conditions
This code is hard to read and maintain

nazarmatsevych (approved):
Nicely done!

PR #79, opened 2023-01-29
Solution

shyptia:


src/generateNumber.js
```diff
@@ -0,0 +1,24 @@
...
+function generateNumber() {
+  const generatedDigits = [];
+
+  const min = 0;
+  const max = 9;
+
+  while (generatedDigits.length !== 4) {
+    const generatedDigit = Math.floor(Math.random() * (max - min + 1)) + min;
+
+    if (generatedDigits[0] === undefined && generatedDigit === 0) {
```
Oleksandr-Filo, line 12:
```suggestion
    if (!generatedDigits[0] && generatedDigit === 0) {
```

src/checkForDublicates.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+function checkForDublicates(number) {
```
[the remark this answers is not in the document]
solaryasha: I agree, it would be more efficient

src/game.js
```diff
@@ -0,0 +1,38 @@
...
+const { calculateBullsAndCows } = require('./calculateBullsAndCows');
+
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+function game(message, generatedNumber) {
+  terminal.question(message, (guessedNumber) => {
+    if (guessedNumber.length !== 4 || checkForDublicates(guessedNumber)) {
```
solaryasha, line 14: What is this number actually mean?
It would help if you created a specific variable for such case

solaryasha (approved):
👍

PR #80, opened 2023-01-30
solution

matvii1:


src/app.js
```diff
@@ -1 +1,52 @@
...
+    const cows = getCows(digitsForGame, userInput);
+
+    if (+bulls === 4) {
+      terminal.write(`You guessed the number in ${moveCounter} moves`);
+      process.exit(1);
+    }
+
+    if (!isFirstZero && !isWrongInput) {
+      moveCounter++;
+      console.log('bulls:', bulls, 'cows:', cows);
```
polosanya, line 37: Try not to use console log. Just return the string instead

src/utils/getRandomNumber.js
```diff
@@ -0,0 +1,23 @@
+function getRandomNumber() {
```
polosanya, line 1: image

src/utils/isNumber.js
```diff
@@ -0,0 +1,17 @@
+'use stric';
```
polosanya, line 1:
```suggestion
'use strict';
```

polosanya (changes requested):
Let's try to pass the tests

src/app.js
matvii1: it does not work this way since return statement stops the function and gets out of it, but I need to continue the game

src/utils/isNumber.js
polosanya: Not fixed

polosanya (changes requested):
Please fix these problems. At least, let's try to disable eslint
image

slavanosatskyi (approved):

polosanya (approved):

PR #81, opened 2023-02-01
done

Nazar-Brunarskyi:


src/getRandomNum.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports.getRandom = () => {
+  const randomNumber = Math.random() * (9999 - 1000) + 1000;
+
+  return Math.floor(randomNumber);
+};
```
GabrielMaksimovich, line 7: It will not pass the p#1 and p#4, because it must be 4 different digits, but here I can get for example number 4444,or 5544, ect.

src/compareTheNumbers.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const { terminal } = require('./terminal');
+const { findBullsAndCows } = require('./findBullsAndCows');
+const isNumber = require('is-number');
+
+function compareTheNumbers(message, randomNumber) {
+  terminal.question(message, (num) => {
+    const isLengthCorrect = num.length === 4;
```
GabrielMaksimovich, line 9: so here when you check the length you also need to check the uniqueness of all four digits.

GabrielMaksimovich (changes requested):

GabrielMaksimovich (approved):
GJ!

src/getRandomNum.js
```diff
@@ -0,0 +1,16 @@
...
+
+function getRandom() {
+  const randomNumber = Math.random() * (9999 - 1000) + 1000;
+  const floorNum = Math.floor(randomNumber);
+
+  if (isAllDigitsUnique(floorNum)) {
+    return floorNum;
+  }
+
+  return getRandom();
```
ivan-kadykalo, line 13: 🔥

ivan-kadykalo (approved):

ivan-kadykalo (changes requested):
app.js in src folder, while all helpers should be in helpers subfolder
image

ivan-kadykalo (approved):
Good job 🔥

PR #86, opened 2023-02-11
completed

ysrckr:


src/io.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const prompt = require('prompt-sync')();
+
+const question = () => {
+  const numbers = prompt('Enter 4 numbers: ');
+
+  if (numbers.length !== 4 || !numbers.match(/^[0-9]+$/)) {
```
meforoma, line 8: you could cover both evaluations within regex
```suggestion
  if (!numbers.match(/^\d{4}$/)) {
```

src/io.js
```diff
@@ -0,0 +1,20 @@
...
+    // eslint-disable-next-line no-console
+    console.log('You must enter 4 numbers!');
+
+    return question();
+  }
+
+  return numbers.split('').map(Number);
```
meforoma, line 15: 👍🏼

src/generator.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const generateRandomNumber = () => {
+  const random = Math.random() * 10;
```
meforoma, line 4: you could also cover the aspect of duplicate numbers
so that that are no 2222 or 1343 etc.

src/generator.js
```diff
@@ -0,0 +1,22 @@
...
+
+const createRandomArray = () => {
+  const array = [];
+
+  for (let i = 0; i < 4; i++) {
+    array.push(generateRandomNumber());
+  }
+
+  return array;
+};
```
meforoma, line 18: i think you could do `(new Array(4)).fill(generateRandomNumber())`

meforoma: nah, i'm wrong ))
your approach would do good

meforoma (approved):
nice. 
see the few comments of mine.

PR #87, opened 2023-02-11
Solution

vadiimvooo:


src/findBullsAndCows.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const findBullsAndCows = (secret, guess) => {
+  const n = secret.length;
```
Anastasia4sik, line 4: It isn't a very informative variable name

Anastasia4sik (changes requested):
GJ, just one thing to correct

Anastasia4sik (approved):

src/findBullsAndCows.js
```diff
@@ -0,0 +1,37 @@
...
+    if (secretCopy[i] !== '') {
+      const j = guessCopy.indexOf(secretCopy[i]);
+
+      if (j > -1) {
+        secretCopy[i] = '';
+        guessCopy[j] = '';
+        cows++;
+      }
+    }
+  }
```
solaryasha, line 29: Do you need two loops or we can do it in one loop?

src/guessNumber.js
```diff
@@ -0,0 +1,40 @@
...
+  terminal.question(message, (guess) => {
+    if (!isNumberValid(guess)) {
+      guessNumber(secret, messages.error);
+
+      return;
+    }
+
+    const { bulls, cows } = findBullsAndCows(secret, guess);
+
+    if (bulls !== 4) {
```
solaryasha, line 24: What is this number actually mean?
It would help if you created a specific variable for such cases

solaryasha (approved):

PR #90, opened 2023-02-13
solution

Fynkzia:


NadiiaKoch (approved):
GJ!

src/guessNumber.js
```diff
@@ -0,0 +1,44 @@
...
+  terminal.question(message, (guess) => {
+    if (!isNumberValid(guess)) {
+      guessNumber(secret, messages.error);
+
+      return;
+    }
+
+    const { bulls, cows } = getBullsAndCows(secret, guess);
+
+    if (bulls !== 4) {
```
solaryasha, line 29: What is this number actually mean?
It would help if you created a specific variable for such cases

solaryasha (approved):

PR #93, opened 2023-02-13
solution

lefteris-L:


src/checkInput.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = (input) => (
+  new Set([...input]).size !== 4
+  || input.length !== 4
+  || [...input].some(d => isNaN(+d))
```
meforoma, line 6: would a simple regex do the job here?
`/^\d{4}$/.test(input)`

meforoma (approved):
sweet 👍🏼 
i'd only do one simplification with the regex - check it out in a comment.

src/checkInput.js
lefteris-L: Nice. This would cover the last 2 cases definitely but we would still need the Set to check for repeated numbers

meforoma: true @lefteris-L 
just combine them in a right way )

PR #96, opened 2023-02-14
completed task

Gintako666:


package.json
```diff
@@ -16,13 +16,17 @@
...
     "eslint": "^5.16.0",
     "eslint-plugin-jest": "^22.4.1",
     "eslint-plugin-node": "^8.0.1",
     "jest": "^24.7.0"
   },
   "mateAcademy": {
     "projectType": "javascript"
+  },
+  "dependencies": {
+    "lodash": "^4.17.21",
```
ivbutsykin, line 29: You don't use `lodash` as I see, so remove this dependency.

src/app.mjs
```diff
@@ -0,0 +1,84 @@
...
+function calculation(arrayNumbers) {
+  let cows = 0;
+  let bulls = 0;
+
+  const alreadySearched = [];
+
+  let iterator = 0;
+
+  while (iterator < 4) {
+    if (arrayNumbers[iterator] === randomNumberFind[iterator]) {
```
ivbutsykin, line 36: `arrayNumbers[iterator]` can we move it separate variable, because I see 6 reuse

ivbutsykin (approved):
Good job!

PR #105, opened 2023-02-20
Implemented Bulls and cows game

viktoriia-kravchenko:


src/bullsCounter.js
```diff
@@ -0,0 +1,33 @@
...
+
+      if (index !== -1) {
+        newSecret.splice(j, 1, 'c');
+        newGuess.splice(index, 1, 'd');
+        cows++;
+      }
+    }
+  }
+
+  return `bulls: ${bulls}, cows: ${cows}`;
```
Cardlice, line 30:
```suggestion
  for (let i = 0; i < 4; i++) {
    if (newSecret[i] === newGuess[i]) {
      bulls++;
    } else if (newGuess.includes(newSecret[i])) {
      cows++;
    }
  }

  return `bulls: ${bulls}, cows: ${cows}`;
```
I don't think you have to do all of this cause you already get a number of 4 different digits

src/guessValidation.js
```diff
@@ -0,0 +1,18 @@
...
+
+const validateGuess = (guess) => {
+  const preparedGuess = guess
+    .split('')
+    .map(el => +el)
+    .sort((a, b) => a - b);
+
+  for (let i = 1; i < preparedGuess.length; i++) {
+    if (preparedGuess[i] === preparedGuess[i - 1]) {
+      console.log('Digits in your number must be unique');
```
Cardlice, line 12: Will be better if you add a few more validations (now you check the uniqueness of the digits of the number, but check if you get a 4-digits number).

src/randomNumGenerator.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+let digitsQuantity = 4;
+
+const getRandomNumber = () => {
+  const randomNumber = [];
+
+  while (digitsQuantity > 0) {
+    const randomDigit = Math.floor(Math.random() * (10 - 0) + 0);
```
Cardlice, line 9:
```suggestion
    const randomDigit = Math.floor(Math.random() * 10);
```

Cardlice (changes requested):

src/bullsCounter.js
viktoriia-kravchenko: Good point! Thanks :)

Cardlice (approved):
GJ!

Ihor-Karpyn (approved):
GJ 🚀

PR #109, opened 2023-02-28
ugly version

nzrii-kysel:


src/game.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+const readline = require('readline');
+const randomizer = require('./randomizer');
```
anna-inozemets, line 4: Maybe it is better to use destruction here and get only needeble function

src/game.js
```diff
@@ -0,0 +1,38 @@
...
+      terminal.question('try again:\n', handleInput);
+    }
+  } catch (error) {
+    terminal.write(`Something is wrong with your input: ${error.message}\n`);
+    terminal.question('try again:\n', handleInput);
+  }
+};
+
+const start = () => {
+  terminal.write('Digits we generated...\n');
```
anna-inozemets, line 34: Maybe it is better to use the sentence which will say that we start a game, and then use your phrase

src/randomizer.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const getDigit = () => Math.floor(Math.random() * 10);
+
+const get4Digits = () => {
+  const digits = [];
+
+  for (let i = 0; i < 4; i++) {
```
anna-inozemets, line 8: I guess, it is possible to remake this logic without nested loops

anna-inozemets (approved):
Nice solution!

src/randomizer.js
nzrii-kysel: agree, I like your version way more

Anna-Tkachenko (approved):
Clear and nice code, Well Done! ⭐️

PR #110, opened 2023-03-02
Develop

SalehAkaJim:


src/modules/counter-of-cows.js
```diff
@@ -0,0 +1,20 @@
...
+
+const countCows = (userInput, number) => {
+  const numberStr = String(number);
+  const inputStr = String(userInput);
+  const count = {
+    bulls: 0,
+    cows: 0,
+  };
+
+  for (let i = 0; i < 4; i++) {
```
meforoma, line 11: why hardcoding `4` - what if you'd like to play with 6 or 9-digit number?

also you'd want to evaluate whether users' input matches the number.length
and probably, to only contain numbers - i'd use regexp for this.

```suggestion
  for (let i = 0; i < numberStr; i++) {
```

src/modules/random-number.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const randomFourDigits = () => Math.floor(1000 + Math.random() * 9000);
```
meforoma, line 3: how about repetitive numbers?
e.g. `0.555 * 9000 + 1000 = 5995` - would be nice so that numbers do not repat
but not crucial

meforoma (changes requested):
please see the comments and let's discuss if you disagree :)

SalehBUD:
> please see the comments and let's discuss if you disagree :)

Thank you Roman 😊, Hopefully they're fixed now

src/app.js
```diff
@@ -12,19 +12,21 @@ let qNumber = 0;
 
 const letsPlay = () => {
   terminal.question(questions[qNumber], (guess) => {
-    if (guess.length !== 4) {
+    const userGuess = String(guess).replace(/\D/g, '');
+
+    if (userGuess.length !== gameNumber.length) {
```
meforoma, line 17: i would just do
```
if (!userGuess.match(/^\d{gameNumber.length}$/)) {
```

meforoma (approved):

PR #111, opened 2023-03-02
Bulls and cows game was realised

anna-inozemets:


src/countBulls.js
```diff
@@ -0,0 +1,21 @@
...
+  let bulls = 0;
+
+  gameNumberItems.forEach((item, index) => {
+    if (item === inputNumberItems[index]) {
+      bulls++;
+    }
+  });
+
+  return bulls;
+}
```
nzrii-kysel, line 17: that would work the same with valid input
```suggestion
function countBulls(inputNumber, gameNumber) {
  const inputNumberItems = inputNumber
    .toString();
  const gameNumberItems = gameNumber
    .toString().split('');
  let bulls = 0;

  gameNumberItems.forEach((item, index) => {
    if (item === inputNumberItems[index]) {
      bulls++;
    }
  });

  return bulls;
}
```

src/getRandomNumber.js
```diff
@@ -0,0 +1,19 @@
...
+
+  while (numbers.length < 4) {
+    const number = Math.floor(Math.random() * 10);
+
+    if (!numbers.includes(number)) {
+      numbers.push(number);
+    }
+  }
+
+  return +numbers.join('');
```
nzrii-kysel, line 14: if you only use the result as a string then why not store it as a string

src/bullsAndCows.js
```diff
@@ -0,0 +1,45 @@
...
+  input: process.stdin,
+  output: process.stdout,
+});
+
+const computerNumber = getRandomNumber();
+
+function handlePlayerNumber(input) {
+  const [bulls, cows] = [
+    countBulls(input, computerNumber),
+    countCows(input, computerNumber),
```
nzrii-kysel, line 18: looks great but I am not sure those two have to be separate functions

Anna-Tkachenko: agree, you can avoid going throw input value 2 times and create one function for calculating both of values

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #112, opened 2023-03-05
Solution

Avichez:
To play the game: 
1) `npm i`
2) `npm run game`

And have fun 🙂


src/countBullsAndCows.js
```diff
@@ -0,0 +1,28 @@
...
+  playerNumber
+    .toString()
+    .split('')
+    .forEach((num, index) => {
+      if (randomNumber.includes(num)) {
+        if (randomNumber.indexOf(num) === index) {
+          countBulls++;
+        } else {
+          countCows++;
+        }
```
OlhaYevstifieieva, line 16: I guess it's better to avoid nested if-conditions here

OlhaYevstifieieva (approved):
Cute and nice work, I like it 💫

src/countBullsAndCows.js
Avichez: Yep agree, I will try to rethink this part.

src/game.js
```diff
@@ -0,0 +1,96 @@
...
+    bullsAndCows();
+
+    return;
+  }
+
+  if (lowerInput === 'no') {
+    terminal.close();
+
+    return;
+  }
```
alexandra-protyanova, line 71: u can use switch here

alexandra-protyanova (approved):
Well done 👍

src/game.js
Avichez: Thank's, I was in doubt here about what is better to use 😅

PR #116, opened 2023-03-06
Develop

Dashaverz:


src/app.js
```diff
@@ -1 +1,37 @@
...
+    throw Error('Please only enter digits.');
+  }
+
+  if (number.length !== 4) {
+    throw Error('Incorrect number of digits.');
+  }
+
+  if (new Set(number).size !== 4) {
+    throw Error('Number should includes 4 different digits.');
+  }
```
vladyslav78292, line 24: check the last point of task, that should be in separated module

src/app.js
```diff
@@ -1 +1,37 @@
...
+
+const { getRandomNumber } = require('./getRandomNumber');
+const { calculateBullsAndCows } = require('./calculateBullsAndCows');
+
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+terminal.question('Guess a number: ', (number) => {
```
vladyslav78292, line 13: i think will be better to create separate function with this logic, and just call it here

src/app.js
```diff
@@ -1 +1,37 @@
...
+  const num = getRandomNumber();
+
+  if (+number === num) {
+    terminal.write('You are a winner!');
+    terminal.close();
+  } else {
+    calculateBullsAndCows(num, number);
+    terminal.write(`The number was ${num}\n`);
+    terminal.write('Try again!');
+    terminal.close();
```
vladyslav78292, line 35: you shouldn`t close terminal if you didnt win

src/getRandomNumber.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+function getRandomNumber() {
+  const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+  let result = '';
+
+  for (let i = 0; i < 4; i++) {
```
vladyslav78292, line 7: good place for while loop

vladyslav78292 (changes requested):
good job

vladyslav78292 (approved):
p.s. i checked this job second time, because it wasn`t saved on platform, that i checked it

src/app.js
alexandra-protyanova: agree

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,41 @@
...
+
+  if (bulls > 1 && cows <= 1) {
+    // eslint-disable-next-line no-console
+    console.log(`${bulls} bulls and ${cows} cow`);
+  }
+
+  if (bulls <= 1 && cows > 1) {
+    // eslint-disable-next-line no-console
+    console.log(`${bulls} bull and ${cows} cows`);
+  }
```
alexandra-protyanova, line 36: the only thing is changing is `s` in the end of the word, create variables with the correct string which depends on amount bulls/cows, to prevent 4 similar `ifs`

src/getRandomNumber.js
alexandra-protyanova: or recursion

alexandra-protyanova (changes requested):
Fix all comments

PR #118, opened 2023-03-06
solution

Lun4trick:


src/app.js
```diff
@@ -1 +1,67 @@
...
+      if (guessArr[ind] === randomNum[ind]) {
+        bulls += 1;
+      } else if (randomNum.includes(num)) {
+        cows += 1;
+      }
+    });
+    console.log(`bulls: ${bulls}, cows: ${cows}`);
+  } else {
+    console.log('enter 4 different numbers!');
+  }
```
naziamoff, line 49: this way it seems clearer and cleaner to me, wdyt?
It is better to avoid nested `ifs` where possible
```suggestion
  if (!guessIsCorrect) {
    console.log('enter 4 different numbers!');
  }

   guessArr.forEach((num, ind) => {
     const isBull = guessArr[ind] === randomNum[ind];
     const isCow = randomNum.includes(num);
     
     if (isBull) {
       bulls += 1;
       return; // for breaking the current iteration of the `forEach`
      } 
      
     if (isCow) {
       cows += 1;
      }
    });
    
    console.log(`bulls: ${bulls}, cows: ${cows}`);
```

src/app.js
```diff
@@ -1 +1,67 @@
...
+    getBullsAndCows(guess);
+
+    if (bulls === 4) {
+      console.log('You win!');
+      input.close();
+    } else {
+      bulls = 0;
+      cows = 0;
+      getGuess();
+    }
```
naziamoff, line 63: same about if/else, it looks way too huge in the code, we can spare some space here
```suggestion
    if (bulls === 4) {
      console.log('You win!');
      input.close();
    } 
    
    bulls = 0;
    cows = 0;
    getGuess();
```

src/app.js
```diff
@@ -1 +1,67 @@
...
+  return result;
+};
+
+const randomNum = getRandomNumber();
+
+console.log(randomNum);
+
+const getBullsAndCows = (guess) => {
+  const guessArr = guess.split('');
+  const guessIsCorrect = guessArr.every((num, ind) => (
```
naziamoff, line 34: from the function name now it seems to me that it makes the guess (cause it has a verb at the beginning). After changing places of two words - it is now clear that we're dealing with a boolean variable
```suggestion
  const isGuessCorrect = guessArr.every((num, ind) => (
```

naziamoff (approved):

PR #130, opened 2023-03-28
solution

kirill-vypirovskyi:


kuznetsova-anastasiia (approved):
GJ

src/calculate.js
```diff
@@ -0,0 +1,22 @@
...
+    if (number[i] === answer[i]) {
+      bulls++;
+      continue;
+    }
+
+    if (number.includes(answer[i])) {
+      cows++;
+      continue;
+    }
+  }
```
illiarerun, line 17: let's use reduce here 💅

src/app.js
```diff
@@ -1 +1,32 @@
...
+const number = generate();
+const question = 'Enter a 4-digit number: ';
+
+const checkAnswer = (message) => {
+  if (isNaN(+message) || message.length !== 4) {
+    console.log('Invalid enter');
+    terminal.question(question, checkAnswer);
+
+    return;
+  }
```
illiarerun, line 17: `the program check the entered number (if it has exactly 4 different digits)`

you I don't see this check

illiarerun (changes requested):

src/app.js
kirill-vypirovskyi: message.length !== 4 ?

> `the program check the entered number (if it has exactly 4 different digits)`
> 
> you I don't see this check

Zelinskyi-Serhii (approved):

src/calculate.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    if (number.includes(char)) {
+      return {
+        ...count,
+        cows: count.cows + 1,
+      };
+    }
+
+    return count;
```
sergii-nosachenko, line 19:
```suggestion
    count.bulls += char === number[index];
    count.cows += (
      char !== number[index] && number.includes(char)
    );

    return count;
```

Use the fact that boolean will be converted to a number when using plus.

sergii-nosachenko (approved):
Good work, commented

PR #135, opened 2023-04-17
solution

Zelinskyi-Serhii:


Oleksandr-kopaevich (approved):
Hey @Zelinskyi-Serhii, the logic is correct, but you need to work on code style improvements and better naming

PR #136, opened 2023-04-18
Solution

timoxa2t:


src/game.js
```diff
@@ -0,0 +1,76 @@
...
+
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+async function game() {
+  myConsole.log('፨ Buls and cows ፨');
+
+  const number = getRandomNumber(1000, 9999).toString();
```
Oleksandr-kopaevich, line 40: Here's you have the next logical bug:
image
All digits in the number must be unique.

src/game.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+async function turn(number) {
+  return new Promise((resolve, reject) => {
+    terminal.question('Gues my 4 digit number: ', (userInput) => {
+      if (userInput === number) {
+        myConsole.log('You won!!!', win);
+        terminal.close();
+        reject(new Error());
```
Oleksandr-kopaevich, line 58: That's good that you used Promise here, but why rejecting when the user wins and then catching this to stop the game? That's confusing... you can resolve with boolean value true/false and then in
`isEnd = await turn(number);`

src/game.js
Oleksandr-kopaevich: Also when I type duplicates in the original game I should get an error that my version contains duplicated digits, but it wasn't specified in the task, so it's up to you to do. But you see from the screenshot that in such case it's easy to guess number, because I just try 
1111
2222
3333
...
and I can see positions for numbers because of word bull  :)

Oleksandr-kopaevich: Also, it's good to check length of what user types

Oleksandr-kopaevich (changes requested):
Hi @timoxa2t, amazing UI, you made a real game, but please fix the bugs in the logic :)

And678 (approved):
Nice, had fun playing, approve :heavy_check_mark:

PR #138, opened 2023-04-21
bulls&cows game created

vladyslav-mosor:


src/app.js
```diff
@@ -1 +1,59 @@
...
+    const guessNumber = () => {
+      terminal.question('\nEnter your guess: ', (guess) => {
+        if (guess.length !== 4) {
+          terminal.write(numberError);
+          guessNumber();
+
+          return;
+        }
+
+        if (!/^[1-4]+$/.test(guess)) {
```
timoxa2t, line 31: Look at example. There aren't only digits from 1 to 4. digits can be from 0 to 9. Otherwise you don't actualy need 'cows' for this game, as it tells you if some digit is in the number

timoxa2t (changes requested):
Good work on code. But you missunderstod the rules of this game.

vladyslav-mosor:
> Good work on code. But you missunderstod the rules of this game.

You're right. It was little misunderstanding with the following example :harold:

valeriiayarzhemska (approved):

And678 (changes requested):
Great work! Your game is really nice.

Unfortunatelly, there is a critical bug hidden here, check out this image:
image

Looks like you regenerate the number on every guess. Make sure the number doesn't change for the entire game.

vladyslav-mosor:
> Great work! Your game is really nice.
> 
> Unfortunatelly, there is a critical bug hidden here, check out this image: image
> 
> Looks like you regenerate the number on every guess. Make sure the number doesn't change for the entire game.

Thank you.
Yes, there were some troubles. Now fixed!

milenadiamonds1 (approved):
Gj! Nice work with attempts left

src/numberGenerator.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const numberGenerator = () => {
+  const number = [];
+
+  while (number.length < 4) {
+    const digit = Math.floor(Math.random() * 9);
```
And678, line 7: Almost fixed, but now this code only generates digits in range 0-8.
Check out [Math.random docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random), `Math.random()` always generates numbers < 1, therefore `Math.random() * 9` will always be < 9.

src/numberGenerator.js
```diff
@@ -0,0 +1,21 @@
...
+
+    if (!number.includes(digit)) {
+      number.push(digit);
+    }
+  }
+
+  if (number[0] === 0) {
+    number[0] = 1;
+  };
```
And678, line 16: Imagine the situation: 

At this point our `number` is `[0,3,1,4]`, we do this check (`if (number[0] === 0) {...`) and **now** the `number` becomes `[1,3,1,4]`, which is incorrect because we have two "1".

Just check inside the `while` loop if we are pushing the first number (you can use `number.length === 0`) and that the new number is not "0".

src/numberGenerator.js
And678: So, the `Math.floor` will never return 9, as your argument is always less than 9.

And678 (changes requested):
Good job fixing the regeneration bug! :fire: 

The number generation is not completely fixed yet, please check my comment below. Feel free to reach out to QnA if you are stuck.

src/numberGenerator.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const numberGenerator = () => {
+  const number = [];
```
And678, line 4: Using a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead of an array would reduce complexity of this generation, because `number.includes` looks through all the items in array until it finds one, Set's `.has` checks for needed item immediately.

```suggestion
  const number = new Set();
```

And678 (changes requested):
Adding to previous review

src/numberGenerator.js
vladyslav-mosor: Finally fixed ✅

And678 (approved):
:muscle:

PR #140, opened 2023-04-26
solution

Deniskakaka:


src/randomNumbers.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = function generationRandomDigits() {
+  let result = '';
+
+  for (let i = 0; i < 4; i++) {
+    result += `${Math.floor(Math.floor(Math.random() * 10))}`;
```
mykhalenych, line 7: Nested Math.floor is unnecessary, use one Math.floor to round the random number.

src/app.js
```diff
@@ -1 +1,57 @@
...
+
+async function playGame() {
+  const numbers = randomDigits();
+  let answear;
+
+  if (!answear) {
+    printMessage('Welcome to Bulls and Cows game!');
+  }
+
+  while (answear !== numbers) {
```
mykhalenych, line 38: It's better to use a different comparison operator like '!==', use 'while (answear.toString() !== numbers.toString())'

mykhalenych (approved):

PR #141, opened 2023-04-26
Solution

valeriiayarzhemska:


src/app.js
```diff
@@ -1 +1,41 @@
+/* eslint-disable no-console */
 'use strict';
+
+const { compareNumbers } = require('./compareNumber');
+const { generateNumber } = require('./generator');
+const { terminal } = require('./terminal');
+
+const generatedNumber = generateNumber();
+
+const checkUserNumber = (userNumber) => {
```
KuharchukDmytro, line 10: rely on task, you need to move it in separate module

image

KuharchukDmytro (approved):
gj!

src/generator.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+function generateNumber() {
+  let generatedNumber = '';
+
+  while (generatedNumber.length < 4) {
+    const randomDigit = (Math.random() * 10).toString().slice(0, 1);
```
Oleksandr-kopaevich, line 7: Your logic is correct, but not reusable and hard to update. I'd take something more general, [like getRandomIntInclusive here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random#getting_a_random_integer_between_two_values)

src/compareNumber.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+function compareNumbers(userNumber, randomNumber) {
+  const bullsAndCows = {
+    bull: 0,
+    cow: 0,
+  };
+
+  for (let i = 0; i < userNumber.length; i++) {
```
Oleksandr-kopaevich, line 9: Try to avoid using old traditional `for` loops. Instead, use specific ones like filter/map/foreach/reduce. Here fits `reduce`.

src/app.js
```diff
@@ -1 +1,41 @@
...
+  const isUserNumberValid = new Set(userNumber).size === 4
```
Oleksandr-kopaevich, line 11: When you use some static values like 4 here especially when you use it in two or more places, it's better to move it to some variable.

src/app.js
```diff
@@ -1 +1,41 @@
...
+
+    return;
+  }
+
+  const { bull, cow } = compareNumbers(userNumber, generatedNumber);
+
+  if (bull === 4) {
+    console.log('U won, congratulations!');
+
+    terminal.close();
```
Oleksandr-kopaevich, line 27: You are missing return statement here, because after showing a user the success message and closing terminal you don't need to execute your code further. 
In your example, your game looks like a bully boy: at first it shows `U won` but then shows `Try again` and just when you are ready to try again it just breaks, without ever letting you to try :)
image

Oleksandr-kopaevich (approved):
Hey @valeriiayarzhemska, your game is working well, I couldn't break it :) I like how clean you made it and the logical flow is easy to follow. I'm approving the work, but please, check my comments to see what can be improved and fixed.

PR #142, opened 2023-04-27
bulls-and-cows game was implemented

KuharchukDmytro:


src/app.js
```diff
@@ -1 +1,6 @@
 'use strict';
+
+const { terminal } = require('./terminal');
+const { handleUserAnswer } = require('./handleUserAnswer');
+
```
And678, line 5: How about adding a welcome message that would be displayed at the start of your game, e.g: "Welcome to bulls and cows".

src/terminal.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const readline = require('readline');
+
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+exports.terminal = terminal;
```
And678, line 10: Instead of exposing the entire readline inteface, export only the methods you need for this app (question, log, etc.)

And678 (changes requested):
Nice! Your game works perfectly, just a couple code related things to improve)

andriibezkorovainyi (approved):
GJ

src/generateNumber.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const generateNumber = () => {
+  let result = '';
+
+  while (result.length < 4) {
```
Oleksandr-kopaevich, line 6: When you use some static values like 4 here especially when you use it in two or more places, it's better to move it to some variable.

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,23 @@
...
+  for (let i = 0; i < userAnswer.length; i++) {
+    if (userAnswer[i] === generatedNumber[i]) {
+      result.bulls += 1;
+      continue;
+    }
+
+    if (generatedNumber.includes(userAnswer[i])) {
+      result.cows += 1;
+    }
+  }
```
Oleksandr-kopaevich, line 18: Try to avoid using old traditional for loops. Instead, use specific ones like filter/map/foreach/reduce. Here fits the best reduce.

Oleksandr-kopaevich (approved):
Hey @KuharchukDmytro, the task is done well, approved!

PR #144, opened 2023-04-27
Bulls and Cows game

Anton-Tretiak:


kateryna-babina (approved):
LGFM

src/app.js
```diff
@@ -1 +1,42 @@
...
+
+const { terminal } = require('./terminal');
+const { generateRandomNumber } = require('./generator');
+const { compareNumbers } = require('./compareNumbers');
+
+const randomNumber = generateRandomNumber();
+
+const playGame = () => {
+  terminal.question('Guess a number: ', (input) => {
+    const formattedInput = input.replace(/\s/g, '');
```
And678, line 12: Good idea removing whitespace :fire:

src/app.js
```diff
@@ -1 +1,42 @@
...
+    const isValidNumber = new Set(formattedInput).size === 4
+      && !isNaN(formattedInput);
+
+    if (!isValidNumber) {
+      console.log('Enter a number consisting of 4 different digits');
```
And678, line 17: Use your terminal module for all input/output. There is a `.write` method in readline ([doc](https://nodejs.org/docs/latest-v18.x/api/readline.html#rlwritedata-key)), or you can make a function that calls console.log in that module.

And678 (changes requested):
Nice work. Code is very clean.
Only one thing to improve below.

src/generator.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const generateRandomNumber = () => {
+  let generatedNumber = '';
```
And678, line 4: For future reference, it would be better to use a `new Set()` here instead of a string (''). You can't add duplicates to set and `set.has(number)` is more efficient than `string.includes(number)`. 

`.includes` searches through every symbol of string one by one, while set's `.has` checks directly.

And678 (approved):
Great work, approve :heavy_check_mark: .
Also, see my comment about using Set below.

PR #145, opened 2023-04-27
Solution

DenysDubas:


src/generateNumber.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+function generateNumber() {
+  const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+  const shuffled = digits.sort(() => Math.random() - 0.5);
```
valeriiayarzhemska, line 5: consider its better to do without sort, what do u think?

valeriiayarzhemska (approved):

src/generateNumber.js
And678, line 5: That's a cool way to do this, but not efficient nor truly random.

`.sort` is a [sorting algorithm](https://blog.shovonhasan.com/time-space-complexity-of-array-sort-in-v8/), and it takes many more comparisons than the number of digits in your array. Checkout the comments for this [stackOverflow answer](https://stackoverflow.com/a/18650169).

And678 (changes requested):
Your solution is clear and concise. Unfortunately it errors because you close your `readline` interface too early.

image

You should close it only after the end of your game.

PR #147, opened 2023-04-27
game done

kateryna-babina:


Yevhenii0536 (approved):

src/app.js
```diff
@@ -1 +1,41 @@
...
+    const { bulls, cows } = calculateBullsAndCows(secretNumber, guess);
+
+    if (bulls === 4) {
+      console.log(`You've found the number
+      ${secretNumber} in ${attempts} attempts.`);
+      terminal.close();
+    } else {
+      console.log(`${bulls} bulls, ${cows} cows`);
+      play();
+    }
```
Oleksandr-kopaevich, line 33: In most cases, it's better to omit the else clause (this comment is related only to code style that can be different on different projects) and have return statements to break further code execution. 

```suggestion
    if (bulls !== 4) {
      console.log(`${bulls} bulls, ${cows} cows`);
      return play();
    }
    
    console.log(`You've found the number
    ${secretNumber} in ${attempts} attempts.`);
    terminal.close();
```

src/app.js
```diff
@@ -1 +1,41 @@
...
+    if (!isValidNumber(guess)) {
+      // eslint-disable-next-line no-console
+      console.log('Wrong data, please enter valid number');
+
+      return play();
+    }
+
+    const { bulls, cows } = calculateBullsAndCows(secretNumber, guess);
+
+    if (bulls === 4) {
```
Oleksandr-kopaevich, line 26: When you use some static values like 4 here especially when you use it in two or more places, it's better to move it to some variable.

src/app.js
```diff
@@ -1 +1,41 @@
...
+      terminal.close();
+    } else {
+      console.log(`${bulls} bulls, ${cows} cows`);
+      play();
+    }
+  });
+}
+
+function isValidNumber(number) {
+  return /^\d{4}$/.test(number) && new Set(number).size === 4;
```
Oleksandr-kopaevich, line 38: RegExp is usually hard to read and it can bring unexpected bugs. A good replacement here can be
`!isNaN(number)`

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+function calculateBullsAndCows(secretNumber, guess) {
+  let bulls = 0;
+  let cows = 0;
+
+  for (let i = 0; i < secretNumber.length; i++) {
```
Oleksandr-kopaevich, line 7: Try to avoid using old traditional for loops. Instead, use specific ones like filter/map/foreach/reduce. Here fits reduce.

Oleksandr-kopaevich (approved):
Hey @kateryna-babina, great job, it works flawlessly. Just see some of my comments about how you can improve the code style.

PR #148, opened 2023-04-27
solution

NikitaStambul:


src/askQuestion.js
```diff
@@ -0,0 +1,33 @@
...
+const { generateDigits } = require('./generateDigits');
+const { terminal } = require('./terminal');
+const conceivedDigits = generateDigits();
+
+function askQueston() {
+  let bulls = 0;
+  let cows = 0;
+
+  terminal.question('Guess the number ', (guessedNumber) => {
+    const guessedDigits = guessedNumber.split('').map(Number);
```
Ambulence, line 13: I think it would be great to write down the rules of the game here. Something like "You need to guess a number with 4 digits

Ambulence (approved):
Great job but it would be cool if the player couldn't enter the wrong number.

src/askQuestion.js
```diff
@@ -0,0 +1,33 @@
...
+
+    for (let i = 0; i < guessedDigits.length; i++) {
+      if (conceivedDigits[i] === guessedDigits[i]) {
+        bulls++;
+      } else if (conceivedDigits.includes(guessedDigits[i])) {
+        cows++;
+      }
+    }
```
And678, line 21: As per the last requirement, move this `calculating bulls and cows` logic to a separate module

And678 (changes requested):
Some things to improve below and this is complete :100:

PR #149, opened 2023-04-27
bulls and cows done

rashyd-hasratov:


Bogdan-Mykhailov (approved):
Cool, good job

src/numberGenerator.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const getFourDigitNumber = () => {
```
Oleksandr-kopaevich, line 3: For now, that's the hardest implementation I've seen for this function related to this task. The second important thing of code after working properly, it's to be friendly to other developers, and these lines are super aggressive. To check it, try to ask some students you have good communication with to tell you what you are trying to do there without showing them the name of the function. Think about how you can simplify it.

src/inputValidator.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const digits = '0123456789';
+
+const isValidInput = (input) => {
+  const hasExactlyFourDigits = input.length === 4;
```
Oleksandr-kopaevich, line 6: When you use some static values like 4 here especially when you use it in two or more places, it's better to move it to some variable.

src/bullsCowsCounter.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const getBullsAndCows = (target, input) => {
+  let bulls = 0;
+  let cows = 0;
+
+  const targetStr = String(target);
+  const inputStr = String(input);
```
Oleksandr-kopaevich, line 8: Don't you get user input as a string already?

Oleksandr-kopaevich (approved):
hey @rashyd-hasratov, the game is working correctly, but please check the code-style comments.

PR #150, opened 2023-04-27
Solution

MaksymMaliuk:


src/generator.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+function generateNumber() {
+  const digits = [];
```
juljaVysotska, line 4: It's excellent idea to use there an array, but it would be better use here `new Set`. [new Set by MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) `Set` could help you store unique values, and you don't need to write additional checking if value exist in your array(9 str in this file).

And in a future, when you need to check if value exist in your `Set` you can use `.has()` method, because it checks directly, than `includes()` that checks values one by one.

src/app.js
```diff
@@ -1 +1,64 @@
...
+  lengthError,
+  rulesMessage,
+  uniqueError,
+  winMessage,
+} = messages;
+
+const randomNumber = generateNumber();
+
+userTerminal.question(helloMessage, (answer) => {
+  if (answer === 'yes') {
```
juljaVysotska, line 21: User can write string `'yes'` like `'Yes'`, `'yeS'`, `'YeS'` and so on. It will be a good idea to bring this string in lowercase, that we correctly understand the answer.

src/app.js
```diff
@@ -1 +1,64 @@
...
+    const game = () => {
+      userTerminal.question(requireMessage, (userNumber) => {
+        if (isNaN(userNumber) || userNumber.length !== 4) {
+          console.log(lengthError);
+          game();
+
+          return;
+        }
+
+        if (userNumber.length !== (new Set(userNumber).size)) {
```
juljaVysotska, line 35: Perfect usage of set!

juljaVysotska (approved):

PR #151, opened 2023-04-27
Solution

Nordexz:
Solution

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,4 +1,5 @@
 'use strict';
+
 const { randomNumberGenerator } = require('./randomNumberGenerator');
 
 randomNumberGenerator();
--- a/src/app.js
+++ b/src/app.js
@@ -1,1 +1,4 @@
 'use strict';
+const { randomNumberGenerator } = require('./randomNumberGenerator');
+
+randomNumberGenerator();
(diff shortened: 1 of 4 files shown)


src/randomNumberGenerator.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const { inputOutputOperation } = require('./inputOutputOperation');
+
+const randomNumberGenerator = () => {
+  const arrOfDigits = [];
+
+  while (arrOfDigits.length !== 4) {
+    const number = Math.floor(Math.random() * 10) + '';
```
And678, line 9: No need to do this.

```suggestion
    const number = Math.floor(Math.random() * 10);
```

src/app.js
```diff
@@ -1 +1,5 @@
 'use strict';
+
+const { randomNumberGenerator } = require('./randomNumberGenerator');
+
+randomNumberGenerator();
```
And678, line 5: `randomNumberGenerator` should not be the implementation of entire game. This violates [sepation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). 

Random number generator should just return a number, no logging, nothing else.
calculateBullsAndCows should not output anything to screen, it should just return something like `{bulls: 0, cows: 2}`.
inputOutput should just export functions for logging and question, no game logic.

And then you need to call all those functions in your main `app.js` to implement a game.

And678 (changes requested):
Game logic works correctly, but needs some refactoring.

PR #153, opened 2023-04-27
created game

milenadiamonds1:


src/app.js
```diff
@@ -1 +1,42 @@
...
+
+const { generateRandomNumber } = require('./generateRandomNumber');
+const { isValid } = require('./isInputValid');
+const { terminal } = require('./terminal');
+const { calculateBullsAndCows } = require('./calculateCowsAndBulls');
+
+const generetadNum = generateRandomNumber();
+
+const checkUserAnswer = (answer) => {
+  if (!isValid) {
```
juljaVysotska, line 12: I think you forgot to call function :)

src/app.js
```diff
@@ -1 +1,42 @@
...
+    console.log('Insert a number of 4 different digits');
```
juljaVysotska, line 13: For all these strings, you can use JSON to store all your text data. Your code will be cleaner and it better for future uses. Because if you want to add some new data or reuse your text message, your data will be in one place, and you can easily access it.

src/app.js
```diff
@@ -1 +1,42 @@
...
+
+    terminal.question(
```
juljaVysotska, line 15: You have duplicated code here and in 36–39 lines. You can use `playGame()` here to avoid duplicates.

src/app.js
```diff
@@ -1 +1,42 @@
...
+    console.log(`Congratulations, you won!`);
+    terminal.close();
+
+    return;
+  }
+  console.log(`${bulls} bulls, ${cows} cows`);
+  playGame();
+};
+
+const playGame = () => {
```
juljaVysotska, line 35: Make sense to relocate this function to ~10 line. Because u can use this function to avoid duplication 15's line.

src/calculateCowsAndBulls.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const calculateBullsAndCows = (randomNumber, userNum) => {
+  return userNum.split('').reduce(
```
juljaVysotska, line 4: Perfect! Brilliant usage of reduce!🔥

src/generateRandomNumber.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function generateRandomNumber() {
```
juljaVysotska, line 3: The logic for generating random numbers is too complicated, since we first get the index, then we take this index from the array, add it to the number, while removing this number from the array. Wouldn't it be easier to just take a randomly generated number and put it in a `Set` until the length of this `Set` is 4?

src/isInputValid.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+const isValid = (input) => {
+  return new Set(input).size === 4
+    && !isNaN(input);
```
juljaVysotska, line 5: Fascinating solution. Good job!

juljaVysotska (changes requested):
Good work! I found some moments, please check and fix them. Best regards💛

And678 (approved):
Nice work, had fun playing, check out my small suggestion below
Approve :heavy_check_mark:

PR #155, opened 2023-04-28
:white_check_mark: implement solution bulls and cows

Bogdan-Mykhailov:
implement solution bulls and cows

### DIFF
--- a/src/calculateAnimals.js
+++ b/src/calculateAnimals.js
@@ -1,0 +1,21 @@
+'use strict';
+
+const calculateAnimals = (generatedNumber, userNumber) => {
+  let bulls = 0;
+  let cows = 0;
+
+  for (let i = 0; i < generatedNumber.length; i++) {
+    if (generatedNumber[i] === userNumber[i]) {
+      bulls++;
+    } else if (generatedNumber.includes(userNumber[i])) {
+      cows++;
+    }
+  }
+
+  return {
+    bulls,
+    cows,
+  };
+};
+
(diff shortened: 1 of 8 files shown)


src/calculateAnimals.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const calculateAnimals = (generatedNumber, userNumber) => {
+  let bulls = 0;
+  let cows = 0;
+
+  for (let i = 0; i < generatedNumber.length; i++) {
```
juljaVysotska, line 7: Here we can use `reduce()` for counting bulls and cows. You leave your data as string to use includes(), but in array we have this method too. Try to do it with reduce().

src/game.js
```diff
@@ -0,0 +1,81 @@
...
+
+console.log(WELCOME);
+
+const playGame = () => {
+  setTimeout(() => {
+    terminal.question(START, (answer) => {
+      if (answer === CONTINUE) {
+        const game = () => {
+          terminal.question(ENTER_THE_NUMBER, (userNumber) => {
+            if (userNumber.length !== 4) {
```
juljaVysotska, line 37: if-else hell 🥲 

Maybe we can do it cleaner and without such a count of nesting?

juljaVysotska (changes requested):
Good job! I found some moments(especially if-else hell), fix it please 🙏
If you have any questions, please ask them in fe_chat and tag mentors.
Best regards 💛

PR #158, opened 2023-04-28
Develop

BozhenaBodnar:


src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const calculateBullsAndCows = (userGuess, secretNumber) => {
+  let bulls = 0;
+  let cows = 0;
+
+  for (let i = 0; i < secretNumber.length; i++) {
```
juljaVysotska, line 7: Here we can use reduce for counting bulls and cows. You leave your data as string to use includes(), but in array we have this method too. Try to do it with reduce().

PR #159, opened 2023-04-28
solution

Lilu580:


src/checkNumbers.js
And678, line 1: `checkNumbers` should only compare the input to generated number and return the results (e.g `{bulls: 1, cows: 2}`). Separate your game code into its own module (`game.js` for example).

src/generateArrayOfRandomNumbers.js
And678, line 1: Its a good style to name your files the same as its main export. So this should be called `generateArrayOfNumbers.js`.

src/checkNumbers.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable no-console */
+'use strict';
+
+const checkNumbers = (
+  terminal, generatedArr
+) => terminal.question('Guess number: ', (number) => {
+  if (String(number).length !== 4) {
```
And678, line 7: Also add check that the input is a number, currently the game considers `asdf` a valid number :)

And678 (changes requested):
Hello! A couple of things to improve:

Check the 4th requirement, you should only accept input if in has 4 **different** digits, otherwise show a message to user.

PR #160, opened 2023-04-29
node bulls and cows

andriibezkorovainyi:


src/app.js
```diff
@@ -1 +1,32 @@
...
+const readline = require('readline');
+
+const { generateNumber } = require('./generateNumber');
+const { checkUsersTry } = require('./checkUsersTry');
+
+const launchGame = () => {
+  const terminal = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
```
And678, line 12: As per requirements, you should have a separate module for input/output. Create it and move your `readline` stuff there.

src/checkUsersTry.js
```diff
@@ -0,0 +1,32 @@
...
+      bulls.push(usersTry[i]);
+    } else if (numberToGuess.includes(usersTry[i])) {
+      cows.push(usersTry[i]);
+    }
+  }
+
+  return `You have
+  bulls: ${bulls.length ? bulls.join(', ') : 'You have no bull for now'}
+  cows: ${cows.length ? cows.join(', ') : 'You have no cows for now'}`;
+}
```
And678, line 30: Don't give away which numbers are bulls/cows. User just needs to know how many bulls and cows are present in user's guess.

src/generateNumber.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+function generateNumber() {
+  const number = new Set();
```
And678, line 4: Good idea using Set :fire:

And678 (changes requested):
Nice work, your solution works flawlessly.
Only a couple of things to improve below is this done :100:

src/checkUsersTry.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  if (usersTry === numberToGuess) {
+    return 'OK';
+  }
+
+  let bulls = 0;
+  let cows = 0;
+
+  for (let i = 0; i < usersTry.length; i++) {
```
Oleksandr-kopaevich, line 19: Try to avoid using old traditional `for` loops. Instead, use specific ones like filter/map/foreach/reduce. Here fits the best `reduce`.

src/checkUsersTry.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+function checkUsersTry(usersTry, numberToGuess) {
+  if (isNaN(+usersTry) || usersTry.length !== 4) {
```
Oleksandr-kopaevich, line 4: You don't need to use the operator `+` to convert `usersTry` to a number because the function isNaN does it inside.

Oleksandr-kopaevich, line 4: When you use some static values like `4` here especially when you use it in two or more places, it's better to move it to some variable.

src/checkUsersTry.js
```diff
@@ -0,0 +1,32 @@
...
+    return 'The number must contain only 4 digits!';
+  }
+
+  if (new Set(usersTry).size !== 4) {
+    return 'The number must contain only unique digits!';
+  }
```
Oleksandr-kopaevich, line 10: The condition here doesn't suit the error message. You are testing length with a static number, that's confusing. The better way to do this is the next:
```suggestion
  if (new Set(usersTry).size !== usersTry.length) {
    return 'The number must contain only unique digits!';
  }
```

src/checkUsersTry.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+function checkUsersTry(usersTry, numberToGuess) {
+  if (isNaN(+usersTry) || usersTry.length !== 4) {
+    return 'The number must contain only 4 digits!';
```
Oleksandr-kopaevich, line 5: This error message doesn't work well with both conditions...
image
I'd split it to two messages:
`isNaN(usersTry)` --> `Please type a valid number using only digits`
`usersTry.length !== 4` --> `The number must contain only 4 digits!`

src/checkUsersTry.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+function checkUsersTry(usersTry, numberToGuess) {
```
Oleksandr-kopaevich, line 3: The name `checkUsersTry` and `usersTry` don't sound well. `checkUserAnswer` and `usersNumber` or `typedNumber` or anything similar.

Oleksandr-kopaevich (approved):
Hey @andriibezkorovainyi, good job, approved. I've left some ideas on how you can improve the code style. Check them, please.

PR #162, opened 2023-05-01
Bulls & Cows game

bohdan-tron:


src/app.js
```diff
@@ -1 +1,32 @@
 'use strict';
+
+const readline = require('readline');
+const { checkInputNumbers } = require('./logic.js');
+const { randomNum } = require('./uniqueNum.js');
```
witflash, line 5: It's better to name the files the same as the modules they return:
```
require('./checkInputNumbers')
require('./randomNum')
```
Of course, if you haven just one export at each one.

witflash (approved):
Good job 👍🏻

src/app.js
bohdan-tron: thanks, got it

PR #163, opened 2023-05-04
Solution

MykhailoIvchenko:


src/generateNumber.js
```diff
@@ -0,0 +1,37 @@
...
+  newNumberAsString += thirdDigit;
+
+  do {
+    fourthDigit = Math.round(Math.random() * 9);
+  } while (newNumberAsString.includes(fourthDigit.toString()));
+
+  newNumberAsString += fourthDigit;
+
+  return newNumberAsString;
+}
```
DarkMistyRoom, line 33: It isn't a reason to reject, but we can do it more simply by using an array with numbers from 0 to 9, randomly choosing one, and removing it from the array

src/calculateCowsAndBulls.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+function calculateCowsAndBulls(generatedNumber, numberFromUser) {
```
DarkMistyRoom, line 3: this function not only calculates the result but also validates the inputted data. Would be better to split this logic

DarkMistyRoom (approved):
Good job

PR #164, opened 2023-05-07
solution

olecsiy-gavrilyuk:


src/randomNumber.js
```diff
@@ -0,0 +1,16 @@
...
+  let number = '';
+  const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+
+  for (let i = 0; i < 4; i++) {
+    const randomIndex = Math.floor(Math.random() * digits.length);
+
+    number += digits.splice(randomIndex, 1);
+  }
+
+  return number + '';
```
And678, line 13: This doesn't change anything
```suggestion
  return number;
```

src/app.js
```diff
@@ -1 +1,34 @@
+/* eslint-disable no-console */
 'use strict';
+
+const readline = require('readline');
+const { bullAndCows } = require('./bull_and_cows');
+
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
```
And678, line 10: Move this to a separate module for input and output operation, as stated in requirements.

src/bull_and_cows.js
```diff
@@ -0,0 +1,31 @@
...
+    if (el === tempArr[i]) {
+      counters.bullCount++;
+    }
+
+    if (userNumArr.includes(el)) {
+      counters.cowCount++;
+    }
+  });
+
+  return { ...counters };
```
And678, line 28: No need to recreate the object here.
```suggestion
  return counters;
```

src/randomNumber.js
```diff
@@ -0,0 +1,16 @@
...
+}
+
+module.exports.generateNumber = generateNumber();
```
And678, line 16: Better export a function than only a single number. You can generate your number at the begging of the game, in app.js.

```suggestion
module.exports = { generateNumber };
```

src/bull_and_cows.js
```diff
@@ -0,0 +1,31 @@
...
+
+const bullAndCows = (userNum) => {
+  const counters = {
+    cowCount: 0,
+    bullCount: 0,
+  };
+  const userNumArr = userNum.split('');
+  const randomNumArr = generateNumber.split('');
+
+  console.log(randomNumArr);
```
And678, line 14: Don't forget to remove testing logs before pushing.
```suggestion
```

And678 (changes requested):
Check this output:
image

Your game not accept invalid input and show messages for:
- Repeating digits
- when the number of digits is not 4
- When other non numbers are in input

Also, for `7986` it counted 2 bulls and **3 cows**. Looks like bulls are additionaly counted as cows too. This is a bug.

PR #166, opened 2023-05-20
Solution

vyurkov:


src/calculatingBullsAndCows.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+function calculatingBullsAndCows(randomNumber, userNumber) {
```
marina-tilniak, line 3: usually verbs are used for function names
```suggestion
function calculateBullsAndCows(randomNumber, userNumber) {
```

src/calculatingBullsAndCows.js
```diff
@@ -0,0 +1,42 @@
...
+  const userNumberArray = userNumber.toString().split('').map(num => +num);
+  const randomNumberArray = randomNumber.toString().split('').map(num => +num);
```
marina-tilniak, line 5: Need to extract this logic into function

src/calculatingBullsAndCows.js
```diff
@@ -0,0 +1,42 @@
...
+    const index = randomNumberArray
+      .findIndex(num => userNumberArray[i] === num);
+
+    if (index >= 0 && index !== i) {
+      cows.push(userNumberArray[i]);
+    }
+  }
+
+  return {
+    status: count === 4,
```
marina-tilniak, line 27: You can use `bulls.length` instead of separate variable here

src/calculatingBullsAndCows.js
```diff
@@ -0,0 +1,42 @@
...
+    if (userNumberArray[i] === randomNumberArray[i]) {
+      count++;
+      bulls.push(userNumberArray[i]);
+      continue;
+    }
+
+    const index = randomNumberArray
+      .findIndex(num => userNumberArray[i] === num);
+
+    if (index >= 0 && index !== i) {
```
marina-tilniak, line 21: Check if `index >= 0` is redundant here. `i` will not be equal to `index` in case `index` is `-1` anyway

src/calculatingBullsAndCows.js
```diff
@@ -0,0 +1,42 @@
...
+    status: count === 4,
+    statusString: ''.concat(
+      bulls.length
+        ? `bull (guessed digits ${bulls.join(' and ')} are on their places) `
+        : '',
+      cows.length
+        ? `cows (digits ${cows.join(' and ')} are present but on wrong places)`
+        : '',
+      !bulls.length && !cows.length ? 'Not a single number guessed' : ''
+    ),
```
marina-tilniak, line 36: Let's extract this logic to separate function and refactor it a bit (e.g. you could return `'Not a single number guessed'` at the beginning). This logic can be simpler

marina-tilniak (changes requested):

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  return ''.concat(
+    bulls.length
+      ? `bull (guessed digits ${bulls.join(' and ')} are on their places) `
+      : '',
+    cows.length
+      ? `cows (digits ${cows.join(' and ')} are present but on wrong places)`
+      : '',
+  );
```
marina-tilniak, line 19:
```suggestion
  let result = '';

  if (bulls.length) {
    result += `bulls (guessed digits ${bulls.join(' and ')} are on their places) `;
  }
  
  if (cows.length) {
    result += `cows (digits ${cows.join(' and ')} are present but on wrong places)`;
  }

  return result;
```

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,51 @@
...
+  for (let i = 0; i < userNumberArray.length; i++) {
+    if (userNumberArray[i] === randomNumberArray[i]) {
+      bulls.push(userNumberArray[i]);
+      continue;
+    }
+
+    const index = randomNumberArray
+      .findIndex(num => userNumberArray[i] === num);
+
+    if (index >= 0 && index !== i) {
```
marina-tilniak, line 38: why didn't you fix this? let's delete extra check

marina-tilniak (approved):

PR #167, opened 2023-05-24
solved

PashaMalyshkin:


src/game.js
```diff
@@ -0,0 +1,20 @@
...
+function playBullsAndCows() {
+  const terminal = createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  terminal.question('Guess the number ', (number) => {
+    console.log(getHint(getRandomNumber(), number));
+    terminal.close();
+  });
```
qoa11a, line 17: With this realisation I have only one try to guess a number. :(

src/hint.js
```diff
@@ -0,0 +1,26 @@
...
+  for (let i = 0; i < secret.length; i++) {
+    if (secret[i] === guess[i]) {
+      bulls++;
+      copySecret = secret.slice(i + 1, secret.length - i);
+    } else {
+      if (copySecret.includes(guess[i]) && !cows.includes(guess[i])) {
+        cows.push(guess[i]);
+      }
+    }
+  }
```
qoa11a, line 21: It's a little bit hard to understand. Try to make it the following way:

Inside your `for` loop compare `guess[i]` and `secret[i]`. If they are equal increment `bulls`. Else, check if `secretArr.includes(guess[i])`. If it includes then increment `cows`.

Then return an object of bulls and cows (it's better to return an object with values because then you can use it to determine if the user has won).

qoa11a (changes requested):
Not bad, but see comments:

liliia-livandovskaya (approved):

PR #172, opened 2023-05-28
Solution

igorkruz:


src/bullsAndCowsCount.js
```diff
@@ -0,0 +1,33 @@
...
+
+  for (let i = 0; i < guessArr.length; i++) {
+    if (guessArr[i] !== 'X' && secretArr.includes(guessArr[i])) {
+      cows++;
+
+      const index = secretArr.indexOf(guessArr[i]);
+
+      secretArr[index] = 'X';
+    }
+  }
```
qoa11a, line 25: It could be done easier. Try to make it the following way:

Inside your `for` loop compare `guessArr[i]` and `secretArr[i]`. If they are equal increment `bulls`. Else, check if `secretArr.includes(guessArr[i])`. If it includes then increment `cows`.

Then return an object of bulls and cows just as you do it.

qoa11a (approved):
Good job! I approve it but I have one suggestion to make your code better.

SerhiiKirik (approved):
Sync 🌍

PR #174, opened 2023-05-29
do you wanna play a game?:smiling_imp:

PavloReutskiy:


src/numberGenerator.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const generateRandomNumber = () => {
+  const min = 1000;
+  const max = 9999;
+
+  const randomNumber = Math.floor(Math.random() * (max - min + 1)) + min;
```
liliia-livandovskaya, line 7: The generated number may have repetitive digits, which is not allowed in Bulls and Cows game. Consider generating a 4-digit number with unique digits.

src/calcBullsAndCows.js
```diff
@@ -0,0 +1,43 @@
...
+    } else {
+      if (secretNumber[i] in countNums) {
+        countNums[secretNumber[i]]++;
+      } else {
+        countNums[secretNumber[i]] = 1;
+      }
+    }
+  }
+
+  for (let i = 0; i < number.length; i++) {
```
liliia-livandovskaya, line 24: Use forEach instead of a for loop to iterate over number

liliia-livandovskaya (approved):

PR #177, opened 2023-05-29
funck works

BondarenkoDmitriy:


src/app.js
```diff
@@ -1 +1,8 @@
 'use strict';
+
+const { randomNumber } = require('./randomNumber');
+const { guessNumber } = require('./guessNumber');
+
+const number = randomNumber();
+
+guessNumber(number);
```
SerhiiKirik, line 8: Consider renaming the 'guessNumber' function to something more descriptive like 'startGame' or 'playGame', as it is not actually guessing the number but rather starting the game.

src/randomNumber.js
```diff
@@ -0,0 +1,21 @@
...
+  let result = '';
+
+  for (let i = 0; i < 10; i++) {
+    digits.push(i);
+  }
+
+  while (result.length < 4) {
+    const randomIndex = Math.floor(Math.random() * digits.length);
+
+    result += digits[randomIndex];
```
SerhiiKirik, line 14: Ensure that the generated number has 4 different digits. The current implementation may produce a number with repeated digits.

src/guessNumber.js
```diff
@@ -0,0 +1,56 @@
...
+
+const readline = require('readline');
+const { checkDigits } = require('./checkDigits');
+
+const inputInterface = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+const isValidGuess = (guess, neededNumber) => {
```
SerhiiKirik, line 12: The `isValidGuess` function should only check if the guess is valid, but should not log any messages. Remove the console.log statements and move them to another function if necessary.

src/checkNumber.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const checkNumber = (checkDigits, imputDigits) => {
```
SerhiiKirik, line 3:
```suggestion
const checkNumber = (checkDigits, inputDigits) => {
```

src/guessNumber.js
SerhiiKirik: Or rename your function

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

SerhiiKirik (approved):
Great 🔥

PR #180, opened 2023-05-30
Solution

yuliia-anisimova:


src/app.js
```diff
@@ -1 +1,38 @@
...
+
+        return playBullsAndCows();
+      }
+
+      const [bulls, cows] = getHint(secret, guess);
+
+      if (bulls === 4) {
+        console.log('You won!');
+        terminal.close();
+      } else {
```
SerhiiKirik, line 31: can you avoid `else / if` construction?

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution 🧑‍💻

SerhiiKirik (approved):
Great 🔥

PR #186, opened 2023-05-30
Solution

DimaDvm:


src/calc.js
```diff
@@ -0,0 +1,25 @@
...
+
+const calc = (userInput, digits) => {
+  const currentNumber = digits;
+  const params = IscalculatingRight(userInput);
+
+  if (params) {
+    let bulls = 0;
+    let cows = 0;
+
+    for (let i = 0; i < currentNumber.length; i++) {
```
liliia-livandovskaya, line 13: Consider using'forEach' instead of traditional 'for' loop for iterating

src/calc.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const { IscalculatingRight } = require('./IscalculatingRight');
+
+const calc = (userInput, digits) => {
+  const currentNumber = digits;
```
liliia-livandovskaya, line 6: The variable 'currentNumber' is not necessary, as you can directly use 'digits' in the rest of the function.

src/random.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const random = () => {
+  const generatedRandom = Math.random().toString().slice(2);
```
liliia-livandovskaya, line 4: The random number generation method may not provide a 4-digit number with unique digits. Consider using a different approach.

src/IscalculatingRight.js
```diff
@@ -0,0 +1,27 @@
...
+  const params = userInput;
+
+  if (params.length !== 4) {
+    throw new Error('Please write 4 different digits');
+  }
+
+  const isUniq = [];
+
+  for (let i = 0; i < params.length; i++) {
+    if (isNaN(+params[i])) {
```
liliia-livandovskaya, line 13: Avoid using unary '+' for type conversion. Use 'Number()'

src/IscalculatingRight.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const IscalculatingRight = (userInput) => {
```
liliia-livandovskaya, line 3: Consider renaming the function 'IscalculatingRight' to something more descriptive and camelCase like 'validateUserInput'.

src/guess.js
```diff
@@ -0,0 +1,38 @@
...
+const rl = require('readline').createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+let tries = 1;
+
+const guess = (digits) => {
+  rl.question(`${tries} try: `, (userInput) => {
+    const bulls = calc(userInput, digits)[0];
```
liliia-livandovskaya, line 15: You are calling the 'calc' function twice, consider storing the result in a variable and use array destructuring to assign values to 'bulls' and 'cows'.

liliia-livandovskaya (approved):
GJ! Pay attention to the comments

PR #192, opened 2023-06-28
Add Bulls and Cows

ruslanpashkov:


src/generateNumber.js
```diff
@@ -0,0 +1,23 @@
...
+  for (let i = 0; i < length; i++) {
+    let number = getRandomDigit();
+
+    while (uniques.includes(number)) {
+      number = getRandomDigit();
+    }
+
+    uniques.push(number);
+  }
+
```
roman-mirzoian, line 17: Now the solution is written correctly, but algorithmically it is not very optimal, because you create a loop (`while`) in the middle of the loop (`for`), instead we can do with one `while` loop with checking values for uniqueness:
```
const wantedNumberLength = 4;
function generateNumber() {
  const uniqueNumbers = [];

  while (uniqueNumbers.length < wantedNumberLength ) {
    const number = getRandomDigit();
    if(!uniqueNumbers.includes(number)) {
      uniqueNumbers.push(number);
    }
  }

  return uniqueNumbers.join('');
}
```

roman-mirzoian (approved):
The work is very well done, I downloaded the application to my computer and even played a little 👍 
I leave the only comment regarding algorithmic optimization, which on small projects does not gain anything in terms of performance, but on large projects can significantly improve performance.

PR #193, opened 2023-06-29
add task solution

NickMitchell61:


src/app.js
```diff
@@ -1 +1,39 @@
...
+      console.log(`Should be a 4-digit number, you entered ${answer.length}\n`);
+      playGame('Try again: \n');
+
+      return;
+    }
+
+    console.log(`Bulls: ${bulls}, Cows: ${cows} \n`);
+
+    if (bulls.length !== number.length) {
+      playGame(`Guess the ${cows.length} digit numbers? \n`);
```
roman-mirzoian, line 29: The message string for starting the game can be placed in a separate variable or function (if you want to dynamically show how many numbers are left to guess) and reuse it.

src/app.js
```diff
@@ -1 +1,39 @@
...
+
+const terminal = rl.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+function playGame(startMessage) {
+  terminal.question(startMessage, (answer) => {
+    const { bulls, cows } = calculating(number, answer);
+
```
roman-mirzoian, line 18: The check for the length of the answer should be before the call to `calculating`. This would prevent unnecessary calculations for invalid inputs.

src/calculating.js
```diff
@@ -0,0 +1,25 @@
...
+
+  const len = answer.toString().length;
+
+  for (let i = 0; i < len; i++) {
+    const pcValue = number.toString().split('')[i];
+    const inputValue = answer.toString().split('')[i];
+
+    if (pcValue === inputValue) {
+      bulls.push(inputValue);
+    } else if (pcValue !== inputValue) {
```
roman-mirzoian, line 15: If you need to check whether `pcValue` is equal or not equal to `inputValue`, it is better to use the `if...else` construction (not `if...else if`).

src/calculating.js
```diff
@@ -0,0 +1,25 @@
...
+      cows.push(inputValue);
```
roman-mirzoian, line 16: This logic only checks whether the digits are the same and in the same position, but does not properly calculate the `cows` according to the task (number of digits in the guessed number that are in the correct number but the wrong place).

src/app.js
```diff
@@ -1 +1,39 @@
+/* eslint-disable no-console */
 'use strict';
+
+const rl = require('readline');
```
roman-mirzoian, line 4: It is better not to shorten the naming to unreadable names, if there are many imports in the project, it will be easy to get confused (`rl` -> `readline`).

src/calculating.js
```diff
@@ -0,0 +1,25 @@
...
+
+const calculating = (number, answer) => {
+  const bulls = [];
+  const cows = [];
+
+  const len = answer.toString().length;
+
+  for (let i = 0; i < len; i++) {
+    const pcValue = number.toString().split('')[i];
+    const inputValue = answer.toString().split('')[i];
```
roman-mirzoian, line 11: If you need to do some operations on the data in the middle of the loop, it's better to move it outside the loop, because now `number.toString().split('')` and `answer.toString().split('')` will be executed on each iteration of the loop (that is, four times), although it is enough to do it once before starting the cycle. This change will improve the performance of the code.

src/app.js
```diff
@@ -1 +1,39 @@
...
+    const { bulls, cows } = calculating(number, answer);
+
+    if (answer.length !== 4) {
+      console.log(`Should be a 4-digit number, you entered ${answer.length}\n`);
+      playGame('Try again: \n');
+
+      return;
+    }
+
+    console.log(`Bulls: ${bulls}, Cows: ${cows} \n`);
```
roman-mirzoian, line 26: Now the validation is going a bit wrong because it show `Cows` even though I haven't guessed any numbers. So the formatting itself is such that nothing is displayed for `Bulls`, it is worth setting the default `0` for these values.
(The first value is that I output the guessed number to the console.)
image

src/calculating.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const calculating = (number, answer) => {
```
roman-mirzoian, line 3: For naming variables and functions, it is better to use names that will tell in more detail what is happening in the code (For example: `calculatingBullsAndCows`, `checkValues` etc. ).

roman-mirzoian (changes requested):
I like the algorithmic approaches that I see in the development of this application 👍 
But there are some questions about the logic itself, the execution of all tasks and optimization of the application.

PR #196, opened 2023-07-09
Bulls and Cows game

Yaroslav-Kononenko:


src/app.js
```diff
@@ -1 +1,27 @@
 'use strict';
+
+const { generateNumber } = require('./numGenerator');
+const { bullsAndCows } = require('./bullsAndCows');
+const { inputFunc } = require('./inputOutput');
```
Anna-Tkachenko, line 5: check function name, let's not make it shortened as it affects in bad way on the code readability

src/app.js
```diff
@@ -1 +1,27 @@
...
+  inputFunc('Enter a 4-digit number: ', (guess) => {
+    if (!isValidGuess(guess)) {
+      game();
+
+      return;
+    }
+
+    const result = bullsAndCows(secretNumber, guess);
+
+    !result.bulls === 4 && game();
```
Anna-Tkachenko, line 19: This code isn't readable
It's the statement for checking if you should call `game` one more time, right? Let's write it with if / else statement

src/app.js
```diff
@@ -1 +1,27 @@
...
+  });
+};
+
+function isValidGuess(guess) {
+  return /^\d{4}$/.test(guess) && new Set(guess).size === 4;
```
Anna-Tkachenko, line 24: Let's create the variable for the template string

Anna-Tkachenko (changes requested):

src/modules/checkIsValidUserInput.js
```diff
@@ -9,7 +9,14 @@
  * @return {boolean} - True if the user input is valid, false otherwise
  */
 function checkIsValidUserInput(userInput) {
-  /* Write your code here */
+  const VALID_GUESS_PATTERN = /^\d{4}$/;
```
BudnikOleksii, line 12: Move constants outside the function

BudnikOleksii (approved):
Good work 👍

PR #198, opened 2023-07-17
made node task bulls-and-cows

vladyslav-lapshyn:


src/app.js
```diff
@@ -1 +1,32 @@
...
+
+    // eslint-disable-next-line no-console
+    console.log(`Bulls: ${bulls}`);
+    // eslint-disable-next-line no-console
+    console.log(`Cows: ${cows}`);
+
+    if (bulls === 4) {
+      guessed = true;
+      // eslint-disable-next-line no-console
+      console.log('Congratulations! You guessed the number!');
```
dinaminahi, line 27: All the messages can be moved so some const variables

dinaminahi (approved):
Well done 👍

src/app.js
vladyslav-lapshyn: @dinaminahi thanks, done

PR #199, opened 2023-07-18
The game has been made

bojkovladislav:


src/generateRandomNumber.js
```diff
@@ -0,0 +1,9 @@
+export const generateRandomNumber = () => {
+  let res = new Set([]);
+
+  while (res.size !== 4) {
+    res.add(Math.floor(Math.random() * 9) + 1);
+  }
+
+  return [...res].join('');
```
loralevitska, line 8: It's better to return an integer or an array of digits for better manipulation instead of a string.
image

src/app.js
```diff
@@ -1 +1,26 @@
...
+  output: process.stdout,
+});
+
+console.log('Welcome to the game bulls and cows!');
+
+const playBullsAndCows = () => {
+  terminal.question('Guess the number ', (userInput) => {
+    calculateBullsAndCows(userInput, randomNumber);
+
+    if (userInput === randomNumber) {
```
loralevitska, line 17: You should use strict comparison '===' to compare values, but you should convert both operands to the same type before comparison. As 'userInput' is a string and 'randomNumber' is a number, they should be converted to the same type before comparison.

src/calculateBullsAndCows.js
```diff
@@ -0,0 +1,39 @@
...
+      res.push('bull')
+      continue;
+    }
+
+    if (updatedGeneratedNumber.includes(updatedUserInput[i])) {
+      res.push('cow')
+    }
+  }
+
+  const numberOfBulls = res.filter(value => value === 'bull').length;
```
loralevitska, line 29: Instead of using `filter` method to calculate the number of bulls and cows, use `reduce` method which would be more efficient for this purpose.

loralevitska (changes requested):
Almost done) Let's improve

loralevitska (approved):
Well done)

PR #200, opened 2023-07-18
solution

ghostargg:


src/app.js
```diff
@@ -1 +1,40 @@
...
+const { check } = require('./checkInput');
+
+const game = createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+const number = generateNumbers();
+
+const runGame = () => {
```
kasiakubisiak, line 15:
```suggestion
const handleRunGame = () => {
```
would be better to start name from `handle`

src/checkInput.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const check = (number, guess) => {
```
kasiakubisiak, line 3:
```suggestion
const handleCheck = (number, guess) => {
```

kasiakubisiak (approved):
I give Approve with just few adjustments, good job!

PR #201, opened 2023-07-18
Solution

skinnyjazz:


src/helper.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const getRandomDigit = (min, max) => {
+  return Math.floor(Math.random() * (max - min + 1) + min);
+};
+
+const getArrayOfrandomValues = (length) => {
```
silchyn, line 7:
```suggestion
const getRandomDigits = (length) => {
```

src/helper.js
```diff
@@ -0,0 +1,22 @@
...
+  while (result.length < length) {
+    if (!result.includes(newValue)) {
+      result.push(newValue);
+    }
+
+    newValue = getRandomDigit(0, 9);
+  }
+
+  return result;
+};
```
silchyn, line 20:
```suggestion
  const result = [];

  for (let i = 0; i < length; ++i) {
    const newValue = getRandomDigit(0, 9);
    
    if (!result.includes(newValue)) {
      result.push(newValue);
    }
  }

  return result;
};
```

src/app.js
```diff
@@ -1 +1,50 @@
...
+
+class Game {
+  constructor() {
+    this.digits = [];
+    this.controller = new GameControl();
+  }
+
+  initialGame() {
+    this.digits = getArrayOfrandomValues(NUMBER_OF_DIGITS);
+    // this.digits = [1, 1, 1, 1];
```
silchyn, line 16:
```suggestion
```

src/app.js
```diff
@@ -1 +1,50 @@
...
+
+    this.controller.onGameStart(
+      NUMBER_OF_DIGITS,
+      this.validateAnswer.bind(this)
```
silchyn, line 20: 🔥

src/app.js
```diff
@@ -1 +1,50 @@
...
+    );
+  }
+
+  validateAnswer(someValues) {
```
silchyn, line 24:
```suggestion
  validateAnswer(answer) {
```

src/gameControl.js
```diff
@@ -0,0 +1,67 @@
...
+
+    this.postQuestion(
+      `Try to guess ${numberOfDigits} numbers: `,
+      validateAnswer
+    );
+  }
+
+  winGame(newGame) {
+    this.terminal.question(
+      'Awsome! You find all bulls! Wont you try again?: yes/no: ',
```
silchyn, line 31:
```suggestion
      'Awesome! You found all bulls! Do you want to try again?: yes/no: ',
```

src/app.js
```diff
@@ -1 +1,50 @@
...
+
+const NUMBER_OF_DIGITS = 4;
+
+class Game {
+  constructor() {
+    this.digits = [];
+    this.controller = new GameControl();
+  }
+
+  initialGame() {
```
silchyn, line 14:
```suggestion
  initGame() {
```

src/app.js
silchyn, line 24: You are not really validating the answer; you're just checking it

silchyn (changes requested):

src/helper.js
skinnyjazz: This doesn't work as intended; in the case of generating an existing number, it will simply return an array with fewer elements.

PR #202, opened 2023-07-18
add solution

growingananas:


src/isGuessValid.js
silchyn, line 1: File could be named `validation`

src/playBullsAndCows.js
```diff
@@ -0,0 +1,43 @@
...
+  input: process.stdin,
+  output: process.stdout,
+});
+
+console.log('Welcome to Bulls and Cows!');
+
+const gameResponse = {
+  error: 'Please enter a valid 4-digit number.',
+  victory: 'Congratulations! You guessed the number',
+};
```
silchyn, line 18: Looks like `enum`

src/app.js
```diff
@@ -1 +1,6 @@
 'use strict';
+
+const { playBullsAndCows } = require('./playBullsAndCows');
+const { generateRandomNumber } = require('./generateRandomNumber');
+
+playBullsAndCows(generateRandomNumber(), 'Enter your 4-digit number: ');
```
silchyn, line 6: This message can also be in a constant

src/playBullsAndCows.js
```diff
@@ -0,0 +1,43 @@
...
+  terminal.question(message, (guess) => {
+    if (!isGuessValid(guess)) {
+      playBullsAndCows(secretNumber, gameResponse.error);
+
+      return;
+    }
+
+    const { bulls, cows } = countBullsAndCows(secretNumber, guess);
+
+    console.log(`Bulls: ${bulls}, Cows: ${cows}. Try again: `);
```
silchyn, line 30: Wait, should you write "try again" if it's a victory?

silchyn (approved):
Good job; try to implement it using classes if you have time

src/playBullsAndCows.js
growingananas: victory is when you guess all 4 digits, nope?

PR #207, opened 2023-07-25
solution

Vadym-Prydatok:


src/calculatingBullsAndCows.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const calculatingBullsAndCows = (secretNumber, inputNumber) => {
+  let cows = 0;
+  let bulls = 0;
+
+  for (let i = 0; i < 4; i++) {
```
OleziO, line 7: Move 4 to the constant variable.

OleziO (changes requested):

PR #208, opened 2023-07-26
sol

GGsmou:


And678 (changes requested):
Doesn't start. Looks like you made a typo.
image

`the program randomly generates a number of 4 **different** digits`
`the player enters a number of 4 **different** digits` - currently i can enter `1111`

And678 (approved):
Good work!

PR #211, opened 2023-08-02
Solution

MihailKostov:


src/modules/validateInput.js
```diff
@@ -0,0 +1,28 @@
...
+    return false;
+  }
+
+  if (value.length !== 4) {
+    console.warn('There should be 4 digits');
+
+    return false;
+  }
+
+  const userNumber = new Set(value);
```
loralevitska, line 16: The value here should be converted to String before applying the Set function. This is because Set function applied directly to number will consider it as single entity and won't check for distinct digits.

src/modules/getBullsAndCows.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const getBullsAndCows = (userInput, randomNumber) => {
```
loralevitska, line 3: The function assumes that the userInput and randomNumber are strings. Consider adding a validation to check if they are indeed strings and if not, convert them.

loralevitska (changes requested):
Almost done) Let's improve

src/modules/getBullsAndCows.js
MihailKostov: I think there is no point in checking randomNumber. The function always returns string

VitaliyBodnar (approved):

PR #212, opened 2023-08-04
solution

MaxSchmide:


src/generator.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const generateRandomNumber = () => {
+  return Math.ceil(Math.random() * 10000);
```
And678, line 4: Not guaranteed to generate 4 **different** digits.

src/app.js
```diff
@@ -1 +1,50 @@
...
+
+const { generateRandomNumber } = require('./generator.js');
+const { terminal } = require('./terminal.js');
+const { check } = require('./checker.js');
+
+const startGame = () => {
+  const number = generateRandomNumber();
+
+  const tryToGuess = () => {
+    terminal.question('Enter 4 digits number: ', (guess) => {
```
And678, line 13:
```suggestion
    terminal.question('Enter a number with 4 different digits: ', (guess) => {
```

And678 (changes requested):
Good job, had fun playing during work :))

Don't forget, you need generate 4 **different** digits, no duplicated. Same for the user input, show `Not valid number!` if it has duplicates

And678 (changes requested):
User input is still not verified, which leads to bugs like this:
image

And678 (approved):
:fire:

PR #214, opened 2023-08-09
solution

cebanoleksandr:


src/generateNumber.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+function generateRandomNumber() {
+  const min = 1000;
+  const max = 9999;
+
+  return Math.floor(Math.random() * (max - min + 1)) + min;
+}
+
+const value = generateRandomNumber().toString();
```
And678, line 10: Export the function, not the number, what if we wanted to play again. The function is called generateNumber.js, not randomNumber.js after all :)

src/generateNumber.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+function generateRandomNumber() {
+  const min = 1000;
+  const max = 9999;
+
+  return Math.floor(Math.random() * (max - min + 1)) + min;
```
And678, line 7: Check the requirements, numbers with duplicate digits are not allowed

And678 (changes requested):
Check requiremens, no duplicate digits in input and generated number.
image


Also, you need a separate module for input/output

And678 (changes requested):
Much better, but some stuff needs to be improved:
Letters are not allowed too:
image

Bulls are not cows (should be 2 bulls and 2 cows here):
image

And678 (changes requested):
Check the image below
image

And678 (approved):
:+1: :+1:

PR #215, opened 2023-08-11
solution

MariiaMonakhova:


package.json
```diff
@@ -16,13 +16,16 @@
...
     "eslint-plugin-jest": "^22.4.1",
     "eslint-plugin-node": "^8.0.1",
     "jest": "^24.7.0"
   },
   "mateAcademy": {
     "projectType": "javascript"
+  },
+  "dependencies": {
+    "readline": "^1.3.0"
   }
```
And678, line 30: Don't install weird modules, readline is part of node

src/app.js
```diff
@@ -1 +1,50 @@
...
+    if (randomNumber === guess) {
+      console.log('Victory!');
+
+      terminal.close();
+
+      return;
+    }
+
+    const bulls = calculateBulls(randomNumber, guess);
+    const cows = calculateCows(randomNumber, guess);
```
And678, line 42: Better merge these to functions to avoid extra loops
```suggestion
    const [bulls, cows] = calculateBullsAndCows(randomNumber, guess);
```

And678 (changes requested):
Letters are not allowed in input, check for that. Also, don't output the solution, ruins the fun.
image

And678 (changes requested):
Letters are still passed. Your should show error on everything **but** the numbers
image

And678 (approved):
:white_check_mark:

PR #216, opened 2023-08-11
add bulls and cows game

ivanbilotserkivskiy:


src/modules/getRandomNumber.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+function getRandomNumber() {
+  return Math.floor(Math.random() * (9999 - 1000) + 1000) + '';
```
And678, line 4: Check the requirements, generated number needs to have unique digits

src/modules/getBullsAndCows.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+function getBullsAndCows(guess, randomNumber) {
+  if (guess.length > 4) {
+    return 'Your number has more than 4 digits';
+  }
+
+  if (guess.length < 4) {
+    return 'Your number has less than 4 digits';
+  }
```
And678, line 10: Better to have validation separate from the calculation of bulls and cows

src/modules/getBullsAndCows.js
```diff
@@ -0,0 +1,37 @@
...
+
+  let bulls = 0;
+  let cows = 0;
+  const target = randomNumber.split('');
+  const testLine = [...target];
```
And678, line 15: No need to make a copy of target, you can safely mutate `target` here
```suggestion
```

And678 (changes requested):
- the player enters a number of 4 **different** digits
- the program randomly generates a number of 4 different digits

Make sure to validate user input to prevent him from entering numbers with duplicate digits

And678 (approved):
:white_check_mark:

PR #217, opened 2023-08-11
Create Bulls and Cows game

kochetov-yevhen-may29:


src/randomNumber.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const fourDigitRandom = Math.floor(1000 + Math.random() * 9000);
```
And678, line 3: Duplicate digits in generated number are not allowed. (check the requirements)

src/randomNumber.js
```diff
@@ -0,0 +1,5 @@
...
+
+module.exports = { fourDigitRandom };
```
And678, line 5: Module should export the function to generate the number, not the number itself

src/checkFuncs.js
And678, line 1: Also check if the input number contains duplicate digits

And678 (changes requested):
Gj, but not all requirements are implemented

src/randomNumber.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+function getFourDigitsRandom() {
+  const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+  let num = '';
+
+  for (let i = 0; i < 4; i++) {
+    const randomIndex = Math.floor(Math.random() * digits.length);
+    num += digits[randomIndex];
+    digits.splice(randomIndex, 1);
```
And678, line 10: Can be a bit simpler
```suggestion
    num += digits.splice(randomIndex, 1)[0];
```

And678 (approved):
:fire:

PR #218, opened 2023-08-11
finished task

artkas03:


src/targetNumber.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const targetNumber = Math.round(Math.random() * 9000 + 999).toString();
+
```
And678, line 4: This will generate duplicate digits:
image

src/targetNumber.js
```diff
@@ -0,0 +1,5 @@
...
+module.exports = { targetNumber };
```
And678, line 5: Better export a function to generate a number, not the number itself, as per requirements:

> Please create separate modules for **generating a number**...

src/terminal.js
```diff
@@ -0,0 +1,19 @@
...
+  input: process.stdin,
+  output: process.stdout,
+});
+
+function askToGuess() {
+  return new Promise((resolve) => {
+    terminal.question('Please enter your guess numeber: ', (number) => {
+      return resolve(number);
+    });
+  });
```
And678, line 13: This is :fire:

src/app.js
```diff
@@ -1 +1,28 @@
...
+    startGame();
+  } else if (userNumber === targetNumber) {
+    terminal.write('\nYou guessed the number! Congratulations!\n');
+
+    terminal.close();
+  } else {
+    const [bulls, cows] = checkBullsAndCows(targetNumber, userNumber);
+
+    terminal.write(
+      `\nNumber of bulls - ${bulls}, number of cows - ${cows} ${targetNumber}\n`
```
And678, line 21: Not fun this way. ;)
```suggestion
      `\nNumber of bulls - ${bulls}, number of cows - ${cows}\n`
```

And678 (changes requested):
Input should not contain duplicate digits as well, show error when it has duplicates. Same for having letters in input:
image

And678 (changes requested):
Good job, approve :white_check_mark:

PR #220, opened 2023-08-14
add solution

IraShtynda:


src/app.js
```diff
@@ -1 +1,50 @@
...
+const checkTurn = (userNumber) => {
+  if (isNaN(userNumber)) {
+    writeResult('You entered invalid number');
+    finishGame();
+
+    return;
+  }
+
+  if (userNumber.length !== NUMBER_OF_DIGITS) {
+    writeResult(`Number should contains ${NUMBER_OF_DIGITS} digits`);
```
And678, line 22: Move these string to constants too: `Number should contains ${NUMBER_OF_DIGITS} digits`)

PR #221, opened 2023-08-14
add task solution

KandaskalovOleg:


And678 (changes requested):
Almost perfect :ok_hand: 
image

PR #223, opened 2023-08-14
add solution

maxyma-sanx:


src/calculator.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable no-console */
+'use strict';
+
+function calculator(userNumbers, generatorNumbers) {
```
anstsot, line 4: better to rename to get more context that function return boolean value

anstsot (approved):

PR #224, opened 2023-08-14
Develop

ihorhladkov:


src/countBullsAndCows.js
```diff
@@ -0,0 +1,26 @@
...
+      } else {
+        cows++;
+      }
+    }
+  }
+
+  return [cows, bulls];
+};
+
+export { countBullsAndCows };
```
anstsot, line 26:
```suggestion
module.exports = {
  countBullsAndCows,
};
```

src/isUniqueArray.js
```diff
@@ -0,0 +1,15 @@
+const isUniqueArray = (array) => {
+  const result = [];
+
+  for (const number of array) {
+    if (!result.includes(number)) {
+      result.push(number);
+    } else {
+      return false;
+    }
```
anstsot, line 9: no need for return
```suggestion
    }
```

anstsot (changes requested):
Try fo change `exports` to `module.exports` that will run the tests on git

src/isUniqueArray.js
ihorhladkov: If I delete this return, I`ll get this.
image

anstsot (approved):

PR #225, opened 2023-08-14
Develop

Ilia991:


src/app.js
```diff
@@ -1 +1,30 @@
...
+  terminalInput.question('Please enter a number ', (number) => {
+    const checkedNumberFromRandom = checkedNumbers(randomNum, number);
+    const checkedBull = checkedNumberFromRandom.filter(item => item === 'bull');
+
+    if (checkedBull.length === 4) {
+      terminalInput.write('You win');
+      terminalInput.close();
+    } else {
+      terminalInput.write('  ');
+      terminalInput.write(checkedNumberFromRandom.join(' '));
```
lerastarynets, line 24: Consider adding a newline character at the end of the printed number for better readability in the terminal.

src/app.js
```diff
@@ -1 +1,30 @@
...
+
+const randomNum = random();
+
+function terminalQuestion() {
+  terminalInput.question('Please enter a number ', (number) => {
+    const checkedNumberFromRandom = checkedNumbers(randomNum, number);
+    const checkedBull = checkedNumberFromRandom.filter(item => item === 'bull');
+
+    if (checkedBull.length === 4) {
+      terminalInput.write('You win');
```
lerastarynets, line 20: Consider adding a newline character at the end of the win message for better readability in the terminal.

src/random.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function random() {
+  let checkedValue = false;
+  const randomNum = Math.ceil(Math.random() * 10000);
+
+  for (let i = 0; i < randomNum.length; i++) {
```
lerastarynets, line 7: randomNum is a number, not a string. You cannot use .length property or indexing on a number. Convert randomNum to a string before this line.

src/random.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function random() {
+  let checkedValue = false;
+  const randomNum = Math.ceil(Math.random() * 10000);
+
+  for (let i = 0; i < randomNum.length; i++) {
+    for (let q = 0; q < randomNum.length; q++) {
+      if (randomNum[i] === randomNum[q]) {
```
lerastarynets, line 9: The if condition will always be true when i equals to q. You need to exclude the case when i equals to q.

src/random.js
```diff
@@ -0,0 +1,19 @@
...
+  for (let i = 0; i < randomNum.length; i++) {
+    for (let q = 0; q < randomNum.length; q++) {
+      if (randomNum[i] === randomNum[q]) {
+        checkedValue = true;
+        break;
+      }
+    }
+  }
+
+  return checkedValue ? random() : String(randomNum);
```
lerastarynets, line 16: In case of checkedValue being true, you are calling random() again. This recursion can risk a stack overflow if a number with unique digits is not found in reasonable iterations.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/random.js
VitaliyBodnar: still not fixed

src/app.js
```diff
@@ -1 +1,29 @@
...
+});
+
+const randomNum = random();
+
+function terminalQuestion() {
+  terminalInput.question('Please enter a number ', (number) => {
+    const checkedNumberFromRandom = checkedNumbers(randomNum, number);
+    const checkedBull = checkedNumberFromRandom.filter(item => item === 'bull');
+
+    if (checkedBull.length === 4) {
```
VitaliyBodnar, line 19: 4 is a 'magic number' so makes sense to move it to the variable

VitaliyBodnar (changes requested):

VitaliyBodnar (changes requested):
.

VitaliyBodnar (changes requested):
.

mykhalenych (approved):

PR #227, opened 2023-08-15
add task solution

irynabolbat:


anstsot (changes requested):
the player enters a number of 4 **different** digits - currently i can enter 1111

Y try to enter one number second time but it automatically add 3 same numbers
image

And678 (approved):
Nice work promisifying terminal.question :fire:

PR #228, opened 2023-08-15
Solution

olena-ole:


anstsot (changes requested):
the player enters a number of 4 **different** digits - currently i can enter 1111

And the text should be in next format: `Cows: 2. Bulls: 1`
image

anstsot (approved):

PR #229, opened 2023-08-15
add task solution

ilona-borkovska:


anstsot (changes requested):
the player enters a number of 4 **different** digits - currently, i can enter `1111`

src/getRandomInt.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function getRandomInt() {
+  const nums = [];
+
+  while (nums.length !== 4) {
+    const randomNum = Math.floor(Math.random() * (9 - 1 + 1) + 1);
```
And678, line 7:
```suggestion
    const randomNum = Math.floor(Math.random() * 9 + 1);
```

And678 (approved):
GJ, approve :white_check_mark:

PR #230, opened 2023-08-16
Solution 1

JohnRust12:


src/app.js
```diff
@@ -1 +1,37 @@
...
+  'Enter 4 different digits: ', (userInput) => {
+    const isValid = validator(userInput);
+
+    if (!isValid) {
+      return request();
+    }
+
+    const result = getBullsAndCows(userInput, randomDigits);
+
+    if (result === 'bull bull bull bull') {
```
lerastarynets, line 26: Checking for 'bull bull bull bull' is not efficient or scalable. Consider returning a count of bulls and checking if it equals 4.

src/check.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+function check(userNum, randomNum) {
```
lerastarynets, line 3: The function should return the number of bulls and cows, not an array with '*' and 'bull' or 'cow'. Consider changing the result format.

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #231, opened 2023-08-17
solution

JOWISSA:


src/generating.js
```diff
@@ -0,0 +1,18 @@
...
+  randomDigits.push(Math.floor(Math.random() * (5 - 3 + 1)) + 3);
+  randomDigits.push(Math.floor(Math.random() * (7 - 6 + 1)) + 6);
+  randomDigits.push(Math.floor(Math.random() * (9 - 8 + 1)) + 8);
+  randomDigits.push(Math.floor(Math.random() * (2 - 0 + 1)) + 0);
+
+  (function shuffle(array) {
+    array.sort(() => Math.random() - 0.5);
+  })(randomDigits);
+
+  return randomDigits.join('');
```
maxim2310, line 15: try to improve you func now its to more redundant lines, also you cant get all digits follow this logic, if you cant make, feel free to ask me in slack

maxim2310 (changes requested):
Almost there) Just one more thing needs to be fixed

PR #232, opened 2023-08-24
Solution

Viktorianeimesh:


.eslintrc.js
```diff
@@ -6,5 +6,5 @@ module.exports = {
   rules: {
     'no-proto': 0
   },
-  plugins: ['jest']
+  plugins: ['jest'],
```
lerastarynets, line 9: Unnecessary trailing comma found in the plugins array. Please remove the comma to adhere to the ESLint rules.

src/numberGenerate.js
```diff
@@ -0,0 +1,17 @@
...
+  const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+
+  for (let i = digits.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1));
+
+    [digits[i], digits[j]] = [digits[j], digits[i]];
+
+    const fourDigitNumber = digits.slice(0, 4).join('');
+
+    return fourDigitNumber;
```
lerastarynets, line 13: The function returns in the first iteration of the loop, so the loop is unnecessary. The return statement should be placed outside the loop.

src/numberGenerate.js
```diff
@@ -0,0 +1,17 @@
...
+
+function generateFourDigitNumber() {
+  const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+
+  for (let i = digits.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1));
+
+    [digits[i], digits[j]] = [digits[j], digits[i]];
+
+    const fourDigitNumber = digits.slice(0, 4).join('');
```
lerastarynets, line 11: The `fourDigitNumber` variable is declared inside the loop and will be reassigned in every iteration. It should be declared once outside the loop.

src/inputAndOutputOperations.js
```diff
@@ -0,0 +1,33 @@
...
+const game = (generatedNumber) => {
+  const terminal = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  terminal.question('Please enter 4 digits: ', (number) => {
+    if (number === generatedNumber) {
+      terminal.write('You won!');
+      terminal.close();
```
lerastarynets, line 16: you can `return` after each condition and just use `if` instead of `else if`

src/inputAndOutputOperations.js
```diff
@@ -0,0 +1,33 @@
...
+    } else if (number.length !== 4) {
+      console.log('The number should consist of 4 digits');
+      terminal.close();
+      game(generatedNumber);
+    } else if (isNaN(number)) {
+      console.log('The number should consist only digits');
+      terminal.close();
+      game(generatedNumber);
+    } else {
```
lerastarynets, line 25: Where is check that all numbers in input are unique?

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/inputAndOutputOperations.js
```diff
@@ -0,0 +1,51 @@
...
+  const terminal = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  terminal.question('Please enter 4 digits: ', (number) => {
+    if (number === generatedNumber) {
+      terminal.write('You won!');
+      terminal.close();
+    }
```
lerastarynets, line 17: why no `return` in the first if block?

lerastarynets (changes requested):
Almost there, just one problem left)

witflash (approved):
Cool! Nice work! 👍🏻

PR #233, opened 2023-08-24
The task is completed

AndruhaMan:


src/generateNumber.js
```diff
@@ -0,0 +1,17 @@
...
+  while (digitsArray.length < NUM_LENGTH) {
+    const numToPush = Math.floor(Math.random() * 10);
+
+    if (!digitsArray.includes(numToPush)) {
+      digitsArray.push(numToPush);
+    }
+  }
+
+  return digitsArray.join('');
+};
```
maxim2310, line 17: you func cant create number with duplicate number and can create number that start from 0, that not so logic

src/validateNumber.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const NUM_LENGTH = 4;
+
+module.exports.isNumberValid = (number) => {
+  for (let i = 1; i < NUM_LENGTH; i++) {
+    if (number[i] === number[i - 1]) {
+      return false;
+    }
```
maxim2310, line 9: this redundant but why do you think that the same number will be placed only together

src/bullsAndCows.js
```diff
@@ -0,0 +1,26 @@
...
+
+    if (isBull) {
+      bullsCount++;
+    } else if (isCow) {
+      cowsCount++;
+    }
+
+    numberToGuessCopy.shift();
+    enteredNumberCopy.shift();
+  }
```
maxim2310, line 23:
```suggestion
  for (let i = 0; i < NUM_LENGTH; i++) {
    const isBull = enteredNumberCopy[0] === numberToGuessCopy[i];
    const isCow = numberToGuess.includes(enteredNumberCopy[i]);

    if (isBull) {
      bullsCount++;
    } else if (isCow) {
      cowsCount++;
    }
  }
```

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

anastasiiavorobiova (approved):
Great work!

PR #240, opened 2023-09-16
bulls_cows

hrynkevych:


src/getBullsAndCows.js
```diff
@@ -0,0 +1,30 @@
...
+
+  if (bulls === 4) {
+    result = true;
+  }
+
+  const resultMessage = result
+    ? 'You won! Congratulations!'
+    : `${bulls} bulls, ${cows} cows`;
+
+  console.log(resultMessage);
```
VitaliyBodnar, line 25: Console logging in the function is not a good practice. It's better to return a result from the function and handle logging elsewhere.

src/getBullsAndCows.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable no-console */
```
VitaliyBodnar, line 1: Disabling eslint is not a good practice, it should be avoided unless it is absolutely necessary. Try to adjust your code to meet the linting standards.

src/app.js
```diff
@@ -1 +1,12 @@
+/* eslint-disable max-len */
```
VitaliyBodnar, line 1: Avoid disabling eslint rules if possible. If the line is too long, consider breaking it into multiple lines or rephrasing it to fit within the max length.

src/game.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable max-len */
+/* eslint-disable no-console */
```
VitaliyBodnar, line 2: You should remove the eslint-disable for max-len. This rule is there to enforce readability and maintainability. Consider refactoring your code to follow this rule.

VitaliyBodnar, line 2: You should remove the eslint-disable for no-console. Instead of using console.log for debugging, consider using a logger or a debugger.

src/game.js
```diff
@@ -0,0 +1,36 @@
...
+
+const game = (
+  message,
+  randomNumber
+) => {
+  terminal.question(
+    message,
+    (guess) => {
+      if (guess.length !== 4
+      || guess.split('').some(num => Number.isNaN(+num))
```
VitaliyBodnar, line 21:
```suggestion
      || guess.split('').some(num => isNaN(+num))
```

src/getRandomNumber.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const getRandomNumber = () => {
+  let numbers = '';
+
+  while (numbers.length < 4) {
+    const number = Math.floor(Math.random() * 10);
```
VitaliyBodnar, line 7: 4 and 10 is a 'magic numbers' so please move them to the constants

VitaliyBodnar (changes requested):

src/getBullsAndCows.js
hrynkevych: thank you for review!

VitaliyBodnar (approved):

PR #244, opened 2023-09-27
Created bulls and cows game

davinnchii:


voievodik (approved):
Good job!
But your random number is generated every time you call the function. I think the main idea of the game is to guess a stable random number. (You just need to subtract the random number from the function)

src/app.js
```diff
@@ -1 +1,33 @@
...
+function gameStart() {
+  terminal.question('Enter a number: ', (userInput) => {
+    if (userInput.length !== 4 || !Number(userInput)) {
+      console.log('Invalid number, please enter a 4 different digits');
+      terminal.question('Enter a number: ', gameStart);
+
+      return;
+    }
+
+    const searchedNumber = generateNumber();
```
ArtemTeslenko, line 17: if user did not win, this number will be generated again and overwrite? In this case it is possible to win only by guessing all the numbers at once.

ArtemTeslenko (changes requested):
Good job! Do not change the source number on each user attempt.

DenisChakhalian (approved):
good job!

src/app.js
```diff
@@ -1 +1,32 @@
...
 'use strict';
+
+const { terminal } = require('./terminal');
+const { generateNumber } = require('./numberGenerator');
+const { checkAnswer } = require('./checkAnswer');
+const searchedNumber = generateNumber();
+
+function gameStart() {
+  terminal.question('Enter a number: ', (userInput) => {
+    if (userInput.length !== 4 || !Number(userInput)) {
```
Moroz-Dmytro, line 11: 4 is a magic value

src/app.js
```diff
@@ -1 +1,32 @@
...
+      console.log('Invalid number, please enter a 4 different digits');
+      terminal.question('Enter a number: ', gameStart);
```
Moroz-Dmytro, line 13:
```suggestion
      gameStart();
```

src/checkAnswer.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+function checkAnswer(searchedNumber, userNumber) {
+  let bulls = 0;
+  let cows = 0;
+
+  const userNumberArr = userNumber.split('');
+
+  for (let i = 0; i < 4; i++) {
```
Moroz-Dmytro, line 9: 4 is a magic value

src/checkAnswer.js
```diff
@@ -0,0 +1,22 @@
...
+    if (searchedNumber.includes(userNumberArr[i])) {
+      if (searchedNumber.indexOf(userNumberArr[i]) === i) {
```
Moroz-Dmytro, line 11: Try not to use nested `if`

src/checkAnswer.js
```diff
@@ -0,0 +1,22 @@
...
+        bulls++;
+      } else {
+        cows++;
+      }
+    }
+  }
+
+  return [bulls, cows];
```
Moroz-Dmytro, line 19: Better to return `object` 
So when you do destructuring order doesn't matter

src/numberGenerator.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+function generateNumber() {
+  return Math.floor(1000 + Math.random() * 9000).toString();
```
Moroz-Dmytro, line 4: Magic values

Moroz-Dmytro (changes requested):

src/app.js
Moroz-Dmytro: Not fixed

src/app.js
```diff
@@ -1 +1,32 @@
...
+    }
+
+    if (userInput === searchedNumber) {
+      console.log('You won!', `Searched number was: ${searchedNumber}`);
+      terminal.close();
+
+      return;
+    }
+
+    const [bulls, cows] = checkAnswer(searchedNumber, userInput);
```
Moroz-Dmytro, line 25: Does it work correctly now? You return object, not array

Moroz-Dmytro (changes requested):
Small question below

Moroz-Dmytro (approved):

PR #245, opened 2023-09-29
Solution

aureliagorzna:
To start the game use command:
`npm run startGame`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,9 +9,12 @@
       "version": "1.0.0",
       "hasInstallScript": true,
       "license": "GPL-3.0",
+      "dependencies": {
+        "prompt-sync": "^4.2.0"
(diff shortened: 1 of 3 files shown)


zaura333 (changes requested):
The rules of the game aren't quite how it's supposed to, according to readme:

> Computer makes `1234`, user prints `1345`. The result is one `bull` (guessed
digit `1` is on it's place) and 2 `cows` (digits `3` and `4` are present but on
wrong places).

It's explained more clearly in current commits to the task. Also the tests are added, so that might be helpful! :)

mbruhler (commented):
true, please refer to readme

mbruhler (changes requested):
That's true, please refer to readme

PR #247, opened 2023-10-02
added logic for bulls and cows game

voievodik:


src/randomizer.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+function getRandomFourDigitNumber() {
+  const min = 1000;
+  const max = 9999;
+
+  return Math.floor(Math.random() * (max - min + 1)) + min;
```
DenisChakhalian, line 7: - the program randomly generates a number of 4 different digits

but in your case - random number can be with the same digits (for ex. 4455 or 6600)

DenisChakhalian (approved):
good job! but check the comment below

ArtemTeslenko (changes requested):
Good job! Couple of issues in comments)

PR #248, opened 2023-10-03
realized Bulls and Cows game with random 4-digit number, user input a…

DenisChakhalian:


src/bullAndCowCounter.js
```diff
@@ -0,0 +1,24 @@
...
+  let cow = 0;
+
+  for (let i = 0; i < arrayOfUserNumbers.length; i++) {
+    if (arrayOfRandomNumbers.includes(arrayOfUserNumbers[i])) {
+      cow++;
+    }
+
+    if (arrayOfRandomNumbers.indexOf(arrayOfUserNumbers[i]) === i) {
+      cow--;
+      bull++;
```
ArtemTeslenko, line 14: a bit of strange approach, instead you can check if array includes, assign to a variable, and then if includes and `i === i` bull++, else cow++

src/app.js
```diff
@@ -1 +1,53 @@
...
+    if (userNumber === randomNumber) {
+      console.log(`You won!`);
+      terminal.close();
+
+      return;
+    }
+
+    const arrayOfUserNumbers = userNumber.split('');
+
+    if (arrayOfUserNumbers.length !== 4 || isNaN(+userNumber)) {
```
roman-mirzoian, line 26: Good message but avoid using "magic values"

src/numberGenerator.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function numberGenerator() {
+  const arrayOfRandomNumbers = [];
+
+  while (arrayOfRandomNumbers.length < 4) {
+    const random = Math.floor(0 + Math.random() * (9 + 1 - 0)).toString();
```
roman-mirzoian, line 7: `0 + Math.random() * (9 + 1 - 0)` - what is this?

src/numberGenerator.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function numberGenerator() {
+  const arrayOfRandomNumbers = [];
+
+  while (arrayOfRandomNumbers.length < 4) {
```
roman-mirzoian, line 6: Avoid using "magic values"

src/numberGenerator.js
```diff
@@ -0,0 +1,19 @@
...
+    const random = Math.floor(0 + Math.random() * (9 + 1 - 0)).toString();
+
+    if (arrayOfRandomNumbers.includes(random)) {
+      continue;
+    }
+
+    arrayOfRandomNumbers.push(random);
+  }
+
+  return arrayOfRandomNumbers.join('');
```
roman-mirzoian, line 16: Why do you use a `join` here and convert the array to a string, so that later in the `app.js` file you call `split` and return it back to the array?

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/numberGenerator.js
BudnikOleksii: > since I need both a string and an array

It's strange behavior in this case, but ok

BudnikOleksii (approved):
Good work 👍 
Approved. Check the comments, I left a few suggestions.

PR #251, opened 2023-10-06
create a workin verion of the game, separate be modules

rostyslav48:
I wanted to make it with imports, but it doesn't work. I would be happy to hear the way I can make it. 
(tried to add `"type": "module"` to package.json, it broke `"require"`, but imports and exports didn't work.)

### DIFF
--- a/src/createTerminal.js
+++ b/src/createTerminal.js
@@ -1,0 +1,14 @@
+'use strict';
+
+const readline = require('readline');
+
+function createTerminal() {
+  return readline.createInterface(
+    process.stdin,
+    process.stdout,
+  );
+};
+
+module.exports = {
+  createTerminal,
+};
--- a/src/getNumbWithRandomDigits.js
+++ b/src/getNumbWithRandomDigits.js
@@ -1,0 +1,24 @@
+'use strict';
+
+function getNumbWithRandomDigits() {
+  const DIGIT_COUNT = 4;
+  const RANDOM_SCOPE = 10;
+
+  let randomDigit = '';
+
+  while ((+randomDigit).toString().length < DIGIT_COUNT) {
+    const newDigit = Math.floor(Math.random() * RANDOM_SCOPE);
+
+    if (randomDigit.includes(`${newDigit}`)) {
+      continue;
+    }
+
+    randomDigit += newDigit;
+  }
+
+  return +randomDigit;
+};
+
+module.exports = {
+  getNumbWithRandomDigits,
+};
--- a/src/processUserGuess.js
+++ b/src/processUserGuess.js
@@ -1,0 +1,32 @@
+'use strict';
+
+const { WINNING_STATUS } = require('./constants');
+
+function processUserGuess(randomNumber, guess) {
+  if (guess === randomNumber) {
+    return WINNING_STATUS;
+  }
+
+  const stringRandomNumber = randomNumber.toString().split('');
+  const stringGuess = guess.toString().split('');
+
+  const isBull = stringRandomNumber
+    .some((numb, index) => +numb === +stringGuess[index]);
+
+  if (isBull) {
+    return 'bull';
+  }
+
+  const isCow = stringRandomNumber
+    .some(numb => stringGuess.includes(numb));
+
+  if (isCow) {
+    return 'cow';
+  }
+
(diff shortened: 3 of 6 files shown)


src/getNumbWithRandomDigits.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+function getNumbWithRandomDigits() {
+  const DIGIT_COUNT = 4;
+  const RANDOM_SCOPE = 10;
+
+  let randomDigit = '';
+
+  while ((+randomDigit).toString().length < DIGIT_COUNT) {
```
Moroz-Dmytro, line 9:
```suggestion
  while ((randomDigit.length < DIGIT_COUNT) {
```
It's not the same?

src/getNumbWithRandomDigits.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+function getNumbWithRandomDigits() {
+  const DIGIT_COUNT = 4;
+  const RANDOM_SCOPE = 10;
```
Moroz-Dmytro, line 5: Move it to file with constants

src/createTerminal.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const readline = require('readline');
+
+function createTerminal() {
```
Moroz-Dmytro, line 5: Why do you need to make it as a function? Why not to just return `terminal` as a constant

src/startGame.js
```diff
@@ -0,0 +1,39 @@
...
+    terminal.close();
+
+    if (number.length !== 4 || isNaN(+number)) {
+      throw new Error('You should input 4 digits');
+    }
+
+    const guessRes = processUserGuess(randomNumber, +number);
+
+    if (guessRes === WINNING_STATUS) {
+      // eslint-disable-next-line
```
Moroz-Dmytro, line 22: You can write `/* eslint-disable no-console */` in the 1 line, not to do it before every console.log

src/processUserGuess.js
```diff
@@ -0,0 +1,32 @@
...
+  if (isBull) {
+    return 'bull';
+  }
+
+  const isCow = stringRandomNumber
+    .some(numb => stringGuess.includes(numb));
+
+  if (isCow) {
+    return 'cow';
+  }
```
Moroz-Dmytro, line 25: image

You should return number of bulls and cows, not just message
For example:
generated number: 1234
user entered number: 2135

You should print message with `bulls: 1, cows: 2`

Moroz-Dmytro (changes requested):

src/getNumbWithRandomDigits.js
rostyslav48: No it is not. I transform the number to int, to avoid 0 as a first number)

src/startGame.js
```diff
@@ -0,0 +1,40 @@
...
+  const randomNumber = numberToGuess || getNumbWithRandomDigits();
+  const terminal = readline.createInterface(
+    process.stdin,
+    process.stdout,
+  );
+
+  terminal.question('Please input your 4 digit number: ', (number) => {
+    terminal.close();
+
+    if (number.length !== 4 || isNaN(+number)) {
```
Moroz-Dmytro, line 19: Magic value 4

src/startGame.js
```diff
@@ -0,0 +1,40 @@
...
+      throw new Error('You should input 4 digits');
```
Moroz-Dmytro, line 20: Better not throw error but notify user that something is wrong and ask question again

src/processUserGuess.js
```diff
@@ -0,0 +1,44 @@
...
+
+  const cowsCount = stringGuess.reduce((acc, numb) => {
+    let increaseBy = 0;
+
+    if (stringRandomNumber.includes(numb)) {
+      increaseBy++;
+    }
+
+    return acc + increaseBy;
+  }, 0);
```
Moroz-Dmytro, line 33: Can we use only 1 loop for this

Moroz-Dmytro (changes requested):

FrontKid (approved):
GOOD JOB!!!

PR #252, opened 2023-10-06
create cli game

neovendano:


src/constants.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const gameNumberSize = 4;
+
+const WRONG_INPUT_MESSAGE
+  = 'You printed something wrong (>_<). Pwease, try again (ﾉ◕ヮ◕)ﾉ';
+const WELCOME_MESSAGE
+  = `Enter ${gameNumberSize} digits pwease (◕‿◕✿): `;
+const VICTORY_MESSAGE
+  = 'Wow, you won! (づ ◕‿◕ )づ';
```
polinavafik, line 10: 🎀✨😍

polinavafik (approved):
great job!

src/constants.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const gameNumberSize = 4;
```
Moroz-Dmytro, line 3: It should be in UPPER_CASE

src/generateFixedSizeNumber.js
```diff
@@ -0,0 +1,19 @@
...
+function generateFixedSizeNumber(length) {
+  let digits = '0123456789';
+  let number = '';
+
+  for (let i = 0; i < length; i++) {
+    const randomIndex = Math.floor(Math.random() * digits.length);
+
+    number += digits[randomIndex];
+    digits = digits.slice(0, randomIndex) + digits.slice(randomIndex + 1);
+  }
```
Moroz-Dmytro, line 12: Would be better if you will generate your number like this `const randomDigit = Math.floor(Math.random() * 10);
`
In this case you don't need additional variable and operations with `slice`, but your solution is interesting 👍

src/app.js
```diff
@@ -1 +1,46 @@
...
+
+      continue;
+    }
+
+    const userNumber = Number(answer);
+
+    if (gameNumber === userNumber) {
+      console.log(VICTORY_MESSAGE);
+      break;
+    } else {
```
Moroz-Dmytro, line 33: Avoid using `else`

Moroz-Dmytro (changes requested):

Moroz-Dmytro (approved):

PR #254, opened 2023-10-07
Implement bulls and cows game

AndriiKuz:


neovendano (approved):
Looking good

src/app.js
```diff
@@ -1 +1,43 @@
...
+      // eslint-disable-next-line no-console
+      console.log(`You won!`);
+      userInterface.close();
+
+      return;
+    }
+
+    if (!isValidDigits(userDigits)) {
+      // eslint-disable-next-line no-console
+      console.log('Invalid digits, try again');
```
neovendano, line 32: You can use constant for console message (at your discretion)

AndriiKuz: Yes, but each of these messages is used only once) So, I think it's ok for this task)

src/modules/generateRandomDigit.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+function generateRandomDigit() {
+  const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+  const result = [];
+
+  const firstDigitIndex = Math.floor(Math.random() * 9) + 1;
+
+  result.push(digits[firstDigitIndex]);
+  digits.splice(firstDigitIndex, 1);
```
Moroz-Dmytro, line 10: Why not to do this inside `for`?
Now you are duplicating part of code

src/modules/calculatingBullsAndCows.js
```diff
@@ -0,0 +1,20 @@
...
+  let cows = 0;
+
+  for (let i = 0; i < computerDigits.length; i++) {
+    if (computerDigits[i] === userDigits[i]) {
+      bulls++;
+    }
+
+    if (computerDigits.includes(userDigits[i])) {
+      cows++;
+    }
```
Moroz-Dmytro, line 14: If you already did `bulls++` you should not do `cows++`

image

src/app.js
```diff
@@ -1 +1,43 @@
...
+  const digitsArr = digits.split('');
+  const uniqueDigits = [...new Set(digitsArr)];
+
+  return uniqueDigits.length === digitsArr.length;
+};
+
+const question = () => {
+  userInterface.question('Enter 4 unique digits: ', (userDigits) => {
+    if (userDigits === computerDigits) {
+      // eslint-disable-next-line no-console
```
Moroz-Dmytro, line 23: You can use `/* eslint-disable no-console */` not to do it before every console.log

Moroz-Dmytro (changes requested):

Moroz-Dmytro (changes requested):

src/modules/generateRandomDigit.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+function generateRandomDigit() {
+  const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+  const result = [];
+
+  for (let i = 0; i < 4; i++) {
```
Moroz-Dmytro, line 7: Here you also have magic value 4 🙂
So you can place your constant in separate file and reuse

PR #257, opened 2023-10-08
Develop

Koliras:


src/app.js
```diff
@@ -1 +1,39 @@
...
+    if (userNumber === randomNumber) {
+      console.log(WIN_MESSAGE);
+      terminal.close();
+
+      return;
+    }
+
+    if (typeof +userInput !== 'number' || userInput.length !== 4) {
+      console.log(INVALID_INPUT_MESSAGE);
+      bullsAndCows();
```
OlhaMomot, line 29: According to the task you are supposed to chek if user entered different numbers -"the program check the entered number (if it has exactly 4 different digits)"

OlhaMomot (approved):
Well done!

src/countBullsAndCows.js
```diff
@@ -0,0 +1,22 @@
...
+    if (generatedNumber.includes(userNumber[i])) {
+      if (generatedNumber.indexOf(userNumber[i]) === i) {
+        bulls += 1;
+      } else {
+        cows += 1;
+      }
+    }
+  }
+
+  return [bulls, cows];
```
Moroz-Dmytro, line 19: Better to return `object` here
So order doesn't matter when you do destructuring

src/countBullsAndCows.js
```diff
@@ -0,0 +1,22 @@
...
+
+  for (let i = 0; i < EXPECTED_NUMBER_LENGTH; i++) {
+    if (generatedNumber.includes(userNumber[i])) {
+      if (generatedNumber.indexOf(userNumber[i]) === i) {
+        bulls += 1;
+      } else {
+        cows += 1;
+      }
+    }
+  }
```
Moroz-Dmytro, line 17: Avoid nested`if`
Try to refactor this part of code

src/app.js
```diff
@@ -1 +1,39 @@
...
+const { generateNumber } = require('./generateNumber');
+const { countBullsAndCows } = require('./countBullsAndCows');
+const {
+  WIN_MESSAGE,
+  INVALID_INPUT_MESSAGE,
+  NUMBER_MESSAGE,
+} = require('./constants');
+
+const randomNumber = generateNumber();
+let userNumber;
```
Moroz-Dmytro, line 14: Why you need this variable?

src/app.js
```diff
@@ -1 +1,39 @@
...
+
+function bullsAndCows() {
```
Moroz-Dmytro, line 16: Function name should start with a verb

src/app.js
```diff
@@ -1 +1,39 @@
...
+      console.log(WIN_MESSAGE);
+      terminal.close();
+
+      return;
+    }
+
+    if (typeof +userInput !== 'number' || userInput.length !== 4) {
+      console.log(INVALID_INPUT_MESSAGE);
+      bullsAndCows();
+    } else {
```
Moroz-Dmytro, line 30: You can avoid using `else` if you use `return`

Moroz-Dmytro (changes requested):
Good work 👍 
Small comments

src/app.js
```diff
@@ -1 +1,42 @@
...
+  INVALID_INPUT_MESSAGE,
+  NUMBER_MESSAGE,
+  EXPECTED_NUMBER_LENGTH,
+} = require('./constants');
+
+const randomNumber = generateNumber();
+
+function playBullsAndCows() {
+  terminal.question(NUMBER_MESSAGE, (userInput) => {
+    const userNumber = userInput;
```
Moroz-Dmytro, line 18: Why you need to create new variable?
Why not to use `userInput`

src/app.js
```diff
@@ -1 +1,42 @@
...
+
+    if (userNumber === randomNumber) {
+      console.log(WIN_MESSAGE);
+      terminal.close();
+
+      return;
+    }
+
+    if (typeof +userInput !== 'number'
+      || [...new Set(userInput.split(''))].length !== EXPECTED_NUMBER_LENGTH) {
```
Moroz-Dmytro, line 28: Create function for this, to check if number is valid
Because to write all this code in `if` is a bad idea

src/countBullsAndCows.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const { EXPECTED_NUMBER_LENGTH } = require('./constants');
+
+function countBullsAndCows(generatedNumber, userNumber) {
+  let bulls = 0;
+  let cows = 0;
+
+  for (let i = 0; i < EXPECTED_NUMBER_LENGTH; i++) {
+    if (generatedNumber.includes(userNumber[i])
```
Moroz-Dmytro, line 10: This check is repeated twice, you can create variable for `generatedNumber.includes(userNumber[i])`
The same with `generatedNumber.indexOf(userNumber[i]) === i`

Moroz-Dmytro (changes requested):
Fix small comments and it's done

src/isInputValid.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+const { EXPECTED_NUMBER_LENGTH } = require('./constants');
+
+function isInputValid(userInput) {
+  return typeof +userInput === 'number'
+    && [...new Set(userInput.split(''))].length === EXPECTED_NUMBER_LENGTH;
```
Moroz-Dmytro, line 7: But would be easier to read if you create variable for this `[...new Set(userInput.split(''))]`

Moroz-Dmytro (approved):
Good work 👍 
CHeck only small comment below

PR #258, opened 2023-10-08
solution🎀

polinavafik:


voievodik (approved):
Such a nice performance 💪

src/constants.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+const MAX_LENGTH = 4;
+
+module.exports = MAX_LENGTH;
```
Moroz-Dmytro, line 5: But better would be to have an object here 🙂

src/getBullsAndCows.js
```diff
@@ -0,0 +1,24 @@
...
+    if (hasCorrectDigit && correctNumber[i] === userNumber[i]) {
+      bulls++;
+    }
+
+    if (hasCorrectDigit && correctNumber[i] !== userNumber[i]) {
+      cows++;
+    }
+  }
+
+  return [bulls, cows];
```
Moroz-Dmytro, line 19: Better to return object `{bulls, cows}`
So the order doesn't matter when you do destructuring

src/app.js
```diff
@@ -1 +1,43 @@
...
+      return;
+    }
+
+    if (isNaN(+userInput) || userInput.length !== 4) {
+      console.log('------------------------------');
+      console.log('I expect a 4 digit num, love');
+      console.log('fix your answer and try again');
+      console.log('------------------------------');
+      startGame();
+    } else {
```
Moroz-Dmytro, line 32: Use `return` to avoid using `else`

Moroz-Dmytro (changes requested):
Great work 🔥 
Really small comments and it's perfect ⭐

Moroz-Dmytro (approved):

PR #259, opened 2023-10-09
add task solution

Wesses:


src/constants.js
```diff
@@ -1,3 +1,5 @@
+'use strict';
+
 const NUMBER_SIZE = 4;
 const MIN_NUMBER = 1000;
```
KatOlista, line 4: number of different digits, it can't be 1000

KatOlista (approved):

src/modules/getBullsCows.js
```diff
@@ -0,0 +1,27 @@
...
+
+  for (const digit of playerNumber) {
+    if (numberToGues.includes(digit)) {
+      cows++;
+    }
+  }
+
+  cows -= bulls;
+
+  return [bulls, cows];
```
Moroz-Dmytro, line 24: Better to return an `object` so order doesn't matter when you do destructuring

src/modules/askForNumber.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+const isNumberValid = require('./isNumberValid');
+const readline = require('readline');
+const getBullsCows = require('./getBullsCows');
+
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
```
Moroz-Dmytro, line 10: image

Moroz-Dmytro (changes requested):

Moroz-Dmytro (approved):

PR #261, opened 2023-10-09
add task solution

Orpheus29:


roman-mirzoian (changes requested):
Very cool done!
One moment left, I entered almost the correct number and repeated the last two digits, but the error was as if I had not entered a four digit number:
image

Koliras (approved):
GJ

PR #262, opened 2023-10-09
implement game cows and bulls

KatOlista:


src/app.js
```diff
@@ -1 +1,45 @@
...
+const { terminal } = require('./terminal');
+const { calculator } = require('./calculator');
+const { output } = require('./output');
+
+const { INPUT_MESSAGE, OUTPUT_MESSAGE } = require('./constants');
+
+const generatedNums = numberGenerator();
+
+const getNumber = (userInput) => {
+  if (userInput.length !== 4) {
```
roman-mirzoian, line 13: Avoid using "magic values".

src/app.js
```diff
@@ -1 +1,45 @@
...
+    terminal.question(
+      INPUT_MESSAGE.enterCorrectNumber,
+      num => getNumber(num)
+    );
+  } else if ((/0-9/).test(userInput)) {
```
roman-mirzoian, line 18: It is better to put this condition in a separate variable to improve readability

src/app.js
```diff
@@ -1 +1,45 @@
...
+    terminal.question(
+      INPUT_MESSAGE.enterCorrectNumber,
+      num => getNumber(num)
+    );
+  } else {
+    const message = output(calculator(generatedNums, userInput));
+
+    if (message === OUTPUT_MESSAGE.win) {
+      // eslint-disable-next-line no-console
```
roman-mirzoian, line 27: If you know that you will need to use many console logs, you can disable it properly for the linter once for the entire file

src/output.js
```diff
@@ -0,0 +1,17 @@
...
+const { OUTPUT_MESSAGE, bullsMessage, cowsMessage } = require('./constants');
+
+const output = ({ bulls, cows }) => {
+  const message = bullsMessage(bulls) + cowsMessage(cows);
+
+  if (!message) {
+    return OUTPUT_MESSAGE.noMatch;
+  } else {
+    return message;
+  }
```
roman-mirzoian, line 12: Try to simplify the code where possible:
```suggestion
  if (!message) {
    return OUTPUT_MESSAGE.noMatch;
  }
  return message;
```

src/constants.js
```diff
@@ -0,0 +1,43 @@
...
+
+const INPUT_MESSAGE = {
+  enterNumber: 'Please enter a number of 4 digits ',
+  enterCorrectNumber: 'Please enter correct number of 4 digits ',
+};
+
+const OUTPUT_MESSAGE = {
+  default: '',
+  noMatch: 'no match',
+  win: 'you won!',
```
roman-mirzoian, line 11:
```suggestion
  win: 'You won!',
```

src/constants.js
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const bullsMessage = (bulls) => {
```
roman-mirzoian, line 14: Why are all sentences in lowercase?

src/constants.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const INPUT_MESSAGE = {
+  enterNumber: 'Please enter a number of 4 digits ',
```
roman-mirzoian, line 4:
```suggestion
  enterNumber: 'Please enter a number of 4 digits: ',
```

src/constants.js
```diff
@@ -0,0 +1,43 @@
...
+  enterCorrectNumber: 'Please enter correct number of 4 digits ',
```
roman-mirzoian, line 5: It is better to indicate that the entered number is incorrect, and not simply to duplicate the previous message with the addition of the word "correct"

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.
Currently, the application does not work exactly as expected in the task.
According to the task, you must return the number of bulls and cows:

image
and you write down which numbers were guessed and how:

image

BudnikOleksii (changes requested):
Good work 👍 
Let's get rid of nested if/else

Moroz-Dmytro (approved):

PR #264, opened 2023-10-13
solution

AndrzejPec:


src/numberGenerator.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports.generateNumber = function() {
+  let digits = '0123456789';
```
anpiet, line 4: I think that digits variable is not necessary. You can multiply Math.random * 9  and have a random digit to use in number variable.
1. task was: the program randomly generates a number of 4 **different** digits. You should generate again random digit if it already exists in number variable.

anpiet (approved):
Just a small change needed. Please find comment attached.
The rest is very good. Well done!

src/numberGenerator.js
AndrzejPec: I actually think this is not the case, because the variable is defined using 'let' so that each time a number is generated, it is also removed from the string, ensuring it will not be picked again, without having to use conditional for discarding a number that has already been generated - which I reckon is what you suggested.

I played the game myself a few times, because it is quite fun after all 😃 And it seems to work fine. Although yeah, it probably could've been accomplished using a different approach.

anpiet: That's true. I haven't checked this deeply. Now I see that you're replacing digit with empty string, so it's not possible to choose the same digit once again.

anpiet (approved):
Everything looks great!

pawel-peksa (approved):
Good job! 🔥

PR #266, opened 2023-10-15
Develop

anpiet:
bulls-and-cows solution

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,8 +1,8 @@
-/* esling-ignore no-console.log */
-"use strict";
+/* eslint-disable no-console */
+'use strict';
 
-const { numGenerator } = require("./numGenerator.js");
-const readline = require("readline");
+const { numGenerator } = require('./numGenerator.js');
+const readline = require('readline');
 
 const digitsToGuess = numGenerator();
 
@@ -10,12 +10,14 @@ const termianl = readline.createInterface({
   input: process.stdin,
   output: process.stdout,
 });
+
 console.clear();
+
 const ask = () => {
   let bull = 0;
   let cow = 0;
 
-  termianl.question("Guess a 4 digit number: ", (num) => {
+  termianl.question('Guess a 4 digit number: ', (num) => {
     for (let i = 0; i < 4; i++) {
       if (num[i] === digitsToGuess[i]) {
         bull++;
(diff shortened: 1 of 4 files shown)


src/numGenerator.js
```diff
@@ -0,0 +1,21 @@
...
+
+  for (let i = 0; i < 4; i++) {
+    const nextDigt = Math.round(Math.random() * 9);
+
+    if (!digits.includes(nextDigt)) {
+      digits += nextDigt;
+    } else {
+      i--;
+    }
+  }
```
AndrzejPec, line 14: In the worst-case scenario, we could spend eternity to generate a number with unique digits because the random number could keep generating numbers that have already been included. Statistically it is rather unlikely, but still this is an area to improve the efficiency, especially in case of scaling the app.

Although it will work, so 🆗

src/app.js
```diff
@@ -1 +1,42 @@
+/* eslint-disable no-console */
 'use strict';
+
+const { numGenerator } = require('./numGenerator.js');
+const readline = require('readline');
+
+const digitsToGuess = numGenerator();
+
+const termianl = readline.createInterface({
```
AndrzejPec, line 9: This typo really needs to be corrected; you really shouldn't let it slide

src/app.js
```diff
@@ -1 +1,42 @@
...
+  output: process.stdout,
+});
+
+console.clear();
+
+const ask = () => {
+  let bull = 0;
+  let cow = 0;
+
+  termianl.question('Guess a 4 digit number: ', (num) => {
```
AndrzejPec, line 20: If my understanding is correct, then there's no validation for inputs of less or more than 4 digits, so the user can put a number of any length, and the game will not complain. While it may not ruin the logic, it sure does make the game somewhat broken.

AndrzejPec (changes requested):
Misspelled variable name must be addressed IMO, I think validation should be added to repair broken mechanism too.

Rest is fine! 👍

src/numGenerator.js
anpiet: Probability that finding last digit for longer than 0.2s is almost equal 0, so tere is no option to run this infinitely.
Random number was often used as a key while mapping array in React. Theoretically all keys could be the same, but probability is almost 0, so it's used.
Task is already accepted by mentor.

src/app.js
pawel-peksa: +1

src/app.js
pawel-peksa: Let's add verification for: 
- length different than 4
- optionally - is the provided input can be coerced to Number

src/numGenerator.js
AndrzejPec: yeah, sure, definitely this is all fine to keep it this way, just wanted to point it out :)

AndrzejPec (approved):

pawel-peksa (approved):
Great Job!

PR #270, opened 2023-11-12
game

OlegPopovych:
Hello,
let's play it!

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.9.7",
+    "@mate-academy/scripts": "^1.2.8",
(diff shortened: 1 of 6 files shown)


src/utils/randomNumberGenerator.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const randomNumberGenerator = () => {
+  const uniqueNumbers = [];
+
+  while (uniqueNumbers.length < 4) {
```
StasSokolov, line 6: let's create variable for `4` and reuse in different places

StasSokolov (changes requested):
Overall looks good, let's fix only one minor comment🔥

src/utils/randomNumberGenerator.js
OlegPopovych: Thanks, now it looks better.

PR #276, opened 2023-11-25
`solution`

sviatikkondur:


src/checkDigits.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (digits.some(digit => isNaN(digit))) {
+    return 'Only numbers are allowed!';
+  }
+
+  if (digits.length !== 4) {
+    return 'You must enter 4 digits';
+  }
+
+  if ([...new Set(digits)].length !== 4) {
```
vasylmiskiv, line 16: Set has a getter `size` and you can do like this
```suggestion
  if (new Set(digits).size !== 4) {
```

PR #280, opened 2023-11-27
implemented task solution

YevhenKharko:


yulabakun (approved):
looks good

src/randomNumber.js
```diff
@@ -0,0 +1,16 @@
...
+  const digits = Array.from({ length: 10 }, (_, i) => i);
+  let result = '';
+
+  for (let i = 0; i < 4; i++) {
+    const randomIndex = Math.floor(Math.random() * digits.length);
+
+    result += digits.splice(randomIndex, 1)[0];
+  }
+
+  return parseInt(result, 10);
```
superpooperxxx, line 13: I think you can generate a number of 4 digits in 1 line

src/prompt.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const readline = require('readline');
+
+const prompt = (msg) => {
```
superpooperxxx, line 5: I cannot se where you validate this
image

superpooperxxx (changes requested):
Looks good, but can be impoved😉

StasSokolov (approved):
### Looks like masterpiece🔥

PR #282, opened 2023-11-28
add solution bulls and cows

Kuper18:
Hi, here is my solution!

### DIFF
--- a/src/generateNum.js
+++ b/src/generateNum.js
@@ -1,0 +1,19 @@
+'use strict';
+
+function getNumbers() {
+  const nums = [];
+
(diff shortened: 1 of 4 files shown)


src/generateNum.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function getNumbers() {
+  const nums = [];
+
+  while (nums.length < 4) {
```
StasSokolov, line 6: not critical but will be good move `4` to the constants and reuse in different places

StasSokolov (approved):
### Looks like masterpiece, i'm approving it with only one small comment🔥

PR #286, opened 2023-12-13
implement solution

kseniia-chepur:


src/app.js
```diff
@@ -1 +1,61 @@
...
+
+    default: {
+      const result = calculateGameResult(randomNumbers, playerInput);
+      displayGameResult(result);
+    }
+  }
+}
+
+function displayGameResult(result) {
+  if (result === true) {
```
VitaliyBodnar, line 50:
```suggestion
  if (result) {
```

src/app.js
```diff
@@ -1 +1,61 @@
...
+const { calculateGameResult } = require('./services/calculateGameResult');
+
+const terminal = readline.createInterface(process.stdin, process.stdout);
+const MIN = 1;
+const MAX = 9;
+const randomNumbers = getRandomNumbers(MIN, MAX);
+
+terminal.question(
+  /* eslint-disable */
+  `In this game you should guess the number. Please enter number of 4 unique digits from ${MIN} to ${MAX} \n`,
```
VitaliyBodnar, line 14: Please split this long text

VitaliyBodnar (changes requested):

src/app.js
kseniia-chepur: I changed the logic slightly, because function displayGameResult in any case returned true or array which also would be considered as true.

Esceype (approved):
Well done!

PR #290, opened 2024-01-04
Develop

Pavlo-Khashchevskyi:


src/modules/createRandomNumber.js
```diff
@@ -0,0 +1,15 @@
+const createRandomNumber = () => {
```
maxim2310, line 1: here you allow to create number that start from 0

src/app.js
```diff
@@ -1 +1,55 @@
...
+const setIsNeedRefresh = (value) => {
+  isNeedRefresh = value;
+};
+
+const setUserNumber = (str) => {
+  userNumber = str;
+  isNeedRefresh = true;
+};
+
+const start = setInterval(() => {
```
maxim2310, line 21: try to rewrite it without interval

maxim2310 (changes requested):
Good work, but lets make some improvement

src/modules/createRandomNumber.js
Pavlo-Khashchevskyi: what is meant? 0123 is this number not correct?

denys-danyliuk: no, number shouldn't start with zero

src/modules/handleTerminal.js
```diff
@@ -0,0 +1,19 @@
+import readline from 'readline';
+
+const handleTerminal = (question, actions) => {
```
denys-danyliuk, line 3: find a better name for this function

denys-danyliuk (approved):

src/modules/createRandomNumber.js
Pavlo-Khashchevskyi: okey i change it. thx

PR #294, opened 2024-01-09
made solution for bullsncows

MrHunterBoi:


src/validator.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const validateGuess = guessNumber => {
```
ksyvvkaa, line 3: you should also check if all digits are different
image


image

src/randomizer.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const getRandomNumber = () => {
```
ksyvvkaa, line 3: number should consist of 4 **_different_** digits
image

ksyvvkaa (changes requested):
well done 😺 but fix comments

MrHunterBoi:
Ah, haven't noticed

src/app.js
```diff
@@ -1 +1,35 @@
...
+  }
+
+  const { bulls, cows } = checkResults(guessNumber, randomNumber);
+
+  console.log(`\nBulls: ${bulls}\nCows: ${cows}\n`);
+
+  if (bulls === 4) {
+    console.log('You guessed the number!');
+    break;
+  }
```
BudnikOleksii, line 34: You can just compare what number user provided with the random number and do it before check bulls and cows

BudnikOleksii (approved):
Good work 👍

PR #298, opened 2024-01-12
done task Bulls and Cows

Anton-Kuchmasov:


package.json
```diff
@@ -16,13 +16,17 @@
...
     "eslint-plugin-jest": "^22.4.1",
     "eslint-plugin-node": "^8.0.1",
     "jest": "^24.7.0"
   },
   "mateAcademy": {
     "projectType": "javascript"
+  },
+  "dependencies": {
+    "readline": "^1.3.0",
+    "readline-sync": "^1.4.10"
```
BudnikOleksii, line 30: You don't need to bring any libraries for this task. And if you do, choose one, you don't need both

src/calculateBullsCows.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  for (let i = 0; i < attempt.length; i++) {
+    if (attempt[i] === rightNumber[i]) {
+      result.bulls++;
+    } else {
+      if (rightNumber.includes(attempt[i])) {
+        result.cows++;
+      }
+    }
```
BudnikOleksii, line 21: Avoid `else` and nested `if` statements

src/app.js
```diff
@@ -1 +1,20 @@
...
+
+const numberToGuess = generateNumber();
+let resultOfGuess = '';
+
+while (true) {
+  resultOfGuess = calculateBullsCows(ioClient(), numberToGuess);
+
+  console.log(resultOfGuess);
+
+  if (resultOfGuess === numberToGuess) {
```
BudnikOleksii, line 16: That looks strange, `calculateBullsCows` returns an object in the format below, but you trying to compare this object with the string `numberToGuess `:
```
const result = {
    bulls: 0,
    cows: 0,
  };
  ```

**Review by BudnikOleksii — CHANGES_REQUESTED** [review]
Good work 👍 
But some things look strange, check the comments

File: src/calculateBullsCows.js [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+  };
+
+  if (attempt.length !== 4) {
+    console.log('Enter a 4-digit number!');
+  }
+
+  if (attempt === rightNumber) {
+    console.log('Congratulations, you won!');
+
+    return -1;
```
BudnikOleksii, line 17: Or `-1` or an `object`, for me it could be a separate function that isn't related to the `calculateBullsCows` or just check.
Because it's strange if the function `calculateBullsCows` doesn't return any bulls/cows 🤷

BudnikOleksii (approved):
👍

PR #301, opened 2024-01-17
the game is created

Anastasiya145:


src/modules/getUserNumber.js
```diff
@@ -0,0 +1,47 @@
...
+    return;
+  }
+
+  for (let i = 0; i < str.length; i++) {
+    if (str.slice(i + 1).includes(str[i])) {
+      console.log('Digits must be different');
+
+      return;
+    }
+  }
```
Andriy-Shymkiv, line 27: you can return false, not just silent return

Andriy-Shymkiv (approved):
lgtm

src/modules/getUserNumber.js
Anastasiya145: ok thanks!

PR #305, opened 2024-01-19
add task solution

thesidorenko:


src/app.js
```diff
@@ -1,3 +1,36 @@
...
+      console.log(chalk.red('Number is not valid. Try again.'));
+      main();
+    } else {
+      const result = getBullsAndCows(input, randomNumber);
+
+      if (result.bulls !== 4) {
+        console.log(`${JSON.stringify(result)}`, chalk.yellow('Guess again'));
+        main();
+      } else {
+        console.log(chalk.green('Congratulations! You guessed the number!'));
```
l4st1m0za, line 29: Consider adding a condition to handle if the input is not a number, for better error handling.

src/modules/getBullsAndCows.js
```diff
@@ -13,7 +13,22 @@
...
-  /* Write your code here */
+  const input = userInput.toString();
+  const number = numberToGuess.toString();
+  const result = {
+    bulls: 0,
+    cows: 0,
+  };
+
+  for (let i = 0; i < input.length; i++) {
+    if (input[i] === number[i]) {
```
l4st1m0za, line 24: The for loop is used here, but it's better to use Array.prototype.forEach for such operations.

l4st1m0za (approved):
Good job, consider changes for the future!

PR #308, opened 2024-01-21
feat: add terminal game functionality

kate-prakofyeva:
(◕ ᴥ ◕) Guess the number of 4 different digits game

### DIFF
--- a/src/modules/checkIsValidUserInput.js
+++ b/src/modules/checkIsValidUserInput.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 'use strict';
 
 /**
@@ -9,9 +10,25 @@
  * @return {boolean} - True if the user input is valid, false otherwise
  */
 function checkIsValidUserInput(userInput) {
-  /* Write your code here */
+  const uniqueNumbers = new Set(userInput);
+
+  if (userInput.length !== 4) {
+    return false;
+  }
+
+  if (isNaN(+userInput)) {
+    return false;
+  }
+
+  if (+userInput < 1000) {
+    return false;
+  }
+
+  if (uniqueNumbers.size !== userInput.length) {
+    return false;
+  }
+
+  return true;
 }
 
-module.exports = {
-  checkIsValidUserInput,
-};
+module.exports = { checkIsValidUserInput };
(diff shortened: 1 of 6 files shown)


src/modules/checkIsValidUserInput.js
```diff
@@ -9,9 +10,25 @@
...
+
+  if (+userInput < 1000) {
+    return false;
+  }
+
+  if (uniqueNumbers.size !== userInput.length) {
+    return false;
+  }
+
+  return true;
```
Patrykmclaren14, line 31: U can write this in one line 

```
  return userInput.length === 4
      && !isNaN(+userInput)
      && +userInput >= 1000
      && new Set(userInput).size === 4;
}
```

kate-prakofyeva: > U can write this in one line
> 
> ```
> function checkIsValidUserInput(userInput) {
>   return userInput.length === 4
>       && !isNaN(+userInput)
>       && +userInput >= 1000
>       && new Set(userInput).size === 4;
> }
> ```

Yes, I can do this, but it seems to me that the option with if is more readable, so it is more optimal from a calculating point of view that you do not have to check all the conditions.

Thank you for your comment 👍 . I would like to know what the mentor thinks about this

mbruhler: This is purely personal and has no performance impact, so whatever (and your team) is up to.
In the professional environment you have to comply to the code styling approved together with your team.

src/modules/getGame.js
```diff
@@ -0,0 +1,50 @@
...
+          console.log(
+            `(◔ ╭╮ ◔) Try again, You found ${bulls} bulls in the right place and ${cows} cows`
+          );
+          getGame(randomNumber);
+        } else {
+          console.log(`(◔ ╭╮ ◔) Try again, You didn't find anything for this shot`);
+          getGame(randomNumber);
+        }
+      }
+    });
```
mbruhler, line 47: Avoid nesting if statements, you can use guards like here https://www.codingbeautydev.com/blog/stop-using-nested-ifs

mbruhler (changes requested):
Please fix your code style

src/modules/getGame.js
kate-prakofyeva: done

mpiskadlo (approved):

PR #309, opened 2024-01-22
Solution

Patrykmclaren14:


Barosz30 (approved):
Yeah, tests work strange here.

Great work! I realy like your code logic and purity

PR #311, opened 2024-01-22
Solution

Barosz30:
🐄🐄🐄🐄


mbruhler (changes requested):
Mooo!
Something is wrong.

Your code uses `const { calculate } = require('./modules/calculate');` which is not in the `modules` directory.

Also please don't skip tests!

PR #313, opened 2024-01-24
Soluuuution

MarcinLigmann:
🐂 & 🐄

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,3 +1,38 @@
+/* eslint-disable no-console */
+/* eslint-disable max-len */
 'use strict';
 
-// Write your code here
+const generateRandomNumber = require('./modules/generateRandomNumber');
+const getBullsAndCows = require('./modules/generateRandomNumber');
+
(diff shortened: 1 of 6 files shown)


src/app.js
```diff
@@ -1,3 +1,38 @@
...
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+const randomNumber = generateRandomNumber();
+
+const game = () => {
+  terminal.question('guess the number from 1234 to 9876 composed of four different numbers', (number) => {
+    if (checkIsValidUserInput(Number(number))) {
```
danon321, line 20: i think smoother will be to check at start 
```
if (!checkIsValidUserInput(Number(number))) {
     console.log('the number you entered is incorrect. ...');
      game();
}
```

and then just do rest of logic. In this case u ommit that one block with if and else

danon321 (changes requested):

danon321 (approved):

PR #318, opened 2024-01-30
Solution

Hladkyi03:


src/modules/generateRandomNumber.js
```diff
@@ -7,7 +7,26 @@
  * @return {number} A random 4-digit number
  */
 function generateRandomNumber() {
-  /* Write your code here */
+  let randomNumber = '';
+  const previousNumbers = [];
+
+  for (let i = 0; i < 4;) {
```
DarkMistyRoom, line 13: why don't you use increment here?

src/modules/generateRandomNumber.js
```diff
@@ -7,7 +7,26 @@
...
+    if (previousNumbers.includes(randomNum)) {
+      continue;
+    }
+
+    previousNumbers.push(randomNum);
+    randomNumber += randomNum.toString();
+    i++;
+  }
+
+  return +randomNumber;
```
DarkMistyRoom, line 29:
```suggestion
  let res = '';
  const numbers = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];

  for (let i = 0; i < 4; i++) {
    const index = Math.floor(Math.random() * (10 - i));

    res += numbers.splice(index, 1)[0];
  }

  return res;
```

src/app.js
```diff
@@ -1,3 +1,48 @@
...
+
+let userInput = null;
+
+function askForNumber() {
+  return new Promise((resolve) => {
+    terminal.question('Enter a 4-digit(unique) number: ', (number) => {
+      if (checkIsValidUserInput(number)) {
+        userInput = number;
+      } else {
+        terminal.write('Invalid input\n');
```
DarkMistyRoom, line 21: would be great to specify what the user has done wrong. You can do it inside the validation function

src/app.js
```diff
@@ -1,3 +1,48 @@
...
+        askForNumber().then(resolve);
```
DarkMistyRoom, line 22: no need to use additional `resolve` here

DarkMistyRoom (changes requested):
Good job! Let's improve it a bit

Hladkyi03:
Your code for generating number dont pass 2 tests, so i didnt use it. Improved validation function, now it gives certain information whats wrong. Also changed a bit asdForNumber function and i need both resolves, if i remove one of them the program will not work with wrong input

anastasiiavorobiova (changes requested):
The previous mentor's comments were not fixed

Hladkyi03:
What didnt fixed? I wrote a message, where explained all details, if u provide a comment u should describe what i need to do

src/app.js
```diff
@@ -1,3 +1,47 @@
...
+let userInput = null;
+
+function askForNumber() {
+  return new Promise((resolve) => {
+    terminal.question('Enter a 4-digit (unique) number: ', (number) => {
+      if (checkIsValidUserInput(number)) {
+        userInput = number;
+        resolve();
+      } else {
+        askForNumber().then(resolve);
```
vadiimvooo, line 22: You should remove redundant then chain

src/modules/generateRandomNumber.js
vadiimvooo, line 29: This block can be modified in the next way
```suggestion
function generateRandomNumber() {
  const secretCode = Array.from({ length: 4 }, (el) => 'x');

  for (let i = 0; i < secretCode.length; i++) {
    secretCode[i] = assignRandomNumber(secretCode);
  }

  return Number(secretCode.join(''));
}

const assignRandomNumber = (secretCode) => {
  const number = Math.floor(Math.random() * 9 + 1);

  if (secretCode.includes(number)) {
    return assignRandomNumber(secretCode);
  }

  return number;
};
```

vadiimvooo (changes requested):
Please check comments and fix the code.

maxim2310 (approved):
Well done

PR #321, opened 2024-02-08
add task solution

DamianAdamczewski:


src/app.js
```diff
@@ -1,3 +1,47 @@
...
+
+const terminal = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+let userInput = null;
+
+async function askForNumber() {
+  while (true) {
```
yevhenii-pyl, line 16: Do we need to introduce a `while(true)` loop here? Let think of a possible refactor here.

yevhenii-pyl (changes requested):
Good job! Let's think how can we avoid while(true) infinite loop :)

PR #327, opened 2024-02-19
add task solution

Softjey:


src/modules/checkIsValidUserInput.js
```diff
@@ -9,7 +13,18 @@
...
  */
 function checkIsValidUserInput(userInput) {
-  /* Write your code here */
+  const notValidLength = userInput.length !== 4;
+  const startWithZero = userInput[0] === '0';
+  const withDuplicates = hasDuplicates(userInput);
+  const notValid = notValidLength || startWithZero || withDuplicates;
+
+  if (notValid) {
+    return false;
```
maxim2310, line 22: would be better to show more informative error for user

Softjey: /**
 * Checks that the user input is valid.
 * Valid user input is a 4-digit number that does not start with 0
@@ -9,7 +13,18 @@
 * @return {boolean} - True if the user input is valid, false otherwise
 */
 
 I followed this instruction. Should this funciton output an error to the console?

lerastarynets (approved):
You're doing amazing, keep it up!

PR #328, opened 2024-02-21
add solution

denys-kononenko:


lerastarynets (approved):
You've got this, keep pushing forward!

PR #364, opened 2024-04-30
game solution added

anastasia1383:


src/modules/generateRandomNumber.js
```diff
@@ -7,7 +7,22 @@
...
+  } while (randomNumber.toString().charAt(0) === '0');
+
+  const digits = randomNumber.toString().split('');
+  const isDuplicate = digits.some(
+    (digit, index) => digits.indexOf(digit) !== index,
+  );
+
+  if (isDuplicate) {
+    return generateRandomNumber();
+  }
```
Moroz-Dmytro, line 23: Try to make up more optimized solution for generating random number, because now this recursion can be called a lot of times

Moroz-Dmytro (approved):

PR #367, opened 2024-05-09
Game solution

oshapkun:


anastasiiavorobiova (changes requested):
Great work! Consider disabling the linter if you need to console log something

PR #373, opened 2024-06-19
Solution

MatheusStefan:


src/modules/getBullsAndCows.js
```diff
@@ -13,7 +13,37 @@
...
+
+      if (!userDigitCount[userStr[i]]) {
+        userDigitCount[userStr[i]] = 0;
+      }
+      guessDigitCount[guessStr[i]]++;
+      userDigitCount[userStr[i]]++;
+    }
+  }
+
+  for (const digit in userDigitCount) {
```
pedro-ruas, line 40: This works, but seems overcomplicated...
Instead of storing values to objects and them iterating, you could use only a single _for loop_ to solve, why don't you give it a try?

pedro-ruas (changes requested):

EdPirro (approved):
Awesome job, @MatheusStefan! Keep it up! 🚀

PR #375, opened 2024-06-20
solution

Allanpt:


src/modules/getBullsAndCows.js
```diff
@@ -12,8 +14,24 @@
...
+  for (let i = 0; i < 4; i++) {
+    if (numberToGuessStrArr.includes(userInputStr[i])) {
+      cows++;
+    }
+
+    if (userInputStr[i] === numberToGuessStrArr[i]) {
+      bulls++;
+      cows--;
+    }
+  }
```
EdPirro, line 32: This could be optimized; currently, its complexity is O(n^2), which is okay in this case because n = 4. However, if we increased the number of digits a guess could have, this function would end up being too inefficient. It is possible to solve this in O(n) with a two-pass solution without too many changes to your implementation (though it would also be possible to achieve a one-pass solution).

As an additional (optional) challenge, could you come up with a more efficient solution?

EdPirro (approved):
Great job, @Allanpt! Keep it up! 🚀

PR #395, opened 2024-08-09
Solution

RostyslavSharuiev:


src/modules/checkIsValidUserInput.js
```diff
@@ -9,7 +9,16 @@
...
+  switch (true) {
+    case userInput.length !== 4:
+    case userInput.startsWith('0'):
+    case new Set(userInput).size !== 4:
+    case Number.isNaN(+userInput):
+      return false;
+
+    default:
+      return true;
+  }
```
VenomMB, line 21:
```suggestion
  return !(userInput.length !== 4 
    && userInput.startsWith('0') 
    && new Set(userInput).size !== 4
    && Number.isNaN(Number(userInput))):
```

src/modules/generateRandomNumber.js
```diff
@@ -7,7 +7,18 @@
  * @return {number} A random 4-digit number
  */
 function generateRandomNumber() {
-  /* Write your code here */
+  const min = 1;
+  const max = 10;
```
VenomMB, line 11:
```suggestion
  const MIN = 1;
  const MAX = 10;
```

VenomMB (approved):
Great job, leave small comments

PR #487, opened 2025-04-17
Develop

yzhyhaliuk:
I implemented validation for both userInput and generateNumber, so there can't be any duplicate digits in the cows calculation. That's what the AI is complaining about.


maxim2310 (approved):
Well done!!!

PR #626, opened 2026-07-19
somution_task_guiderline

Moises-S001:
solution


Moises-S001:
ainda esperando uma resposta

src/modules/generateRandomNumber.js
```diff
@@ -7,7 +7,19 @@
  * @return {number} A random 4-digit number
  */
 function generateRandomNumber() {
-  /* Write your code here */
+  const numerosGerados = [Math.floor(Math.random() * 9) + 1];
```
raulriato: It's not a good practice to mix languages in your code. Here, the code needs to be written in english.

src/app.js
```diff
@@ -1,3 +1,31 @@
...
+      if (resultado.bulls === 4) {
+        rl.close();
+      } else {
+        palpite();
+      }
+    } else {
+      palpite();
+    }
+  });
+}
```
raulriato: You should show the player it's progress on each attempt. something like this example:

Imagine the number generated for the game is `1234`.

The game should be something like this:

```bash
Enter a 4-digits number: 1385 # 1385 is the user input
1 bulls and 1 cows # there is no need to implement the correct use of plural, you can just use fixed bulls and cows.
Enter a 4-digits number: 1432
2 bulls and 2 cows
Enter a 4-digits number: 1234
Congratulations, you got it.

**Review by raulriato — CHANGES_REQUESTED** [review]
The functionality of the game seems correct, but it's not actually playable, since the player does not have any reference if they're close to the correct answer.

Also, make sure to write your code in english. Good luck.

File: src/app.js (hunk 2 above)
Comment (Moises-S001) [reply src/app.js]: para esse caso eu posso usar o 'console.log' ?

File: src/app.js [hunk 3]
```diff
@@ -1,3 +1,35 @@
...
+      if (result.bulls === 4) {
+        process.stdout.write('Congratulations, you got it.');
+        rl.close();
+      } else {
+        prediction();
+      }
+    } else {
+      process.stdout.white('Invalid! Please enter a 4 digits number again.');
+      prediction();
+    }
```
raulriato, line 31: Just a tip for real life code.

instead of using `if` with the "good" conditions and than having to use `else` to deal with the errors, it's much cleaner to treat the problems first:

```js
if (!checkIsValidUserInput(res)) {
  process.stdout.white('Invalid! Please enter a 4 digits number again.');
  prediction();
}

const result = getBullsAndCows(res, numberRandom);

process.stdout.write(`Bulls: ${result.bulls}, Cows: ${result.cows}\n`);

if (result.bulls !== 4) {
  prediction();
  return;
}

process.stdout.write('Congratulations, you got it.');
rl.close();

File: src/app.js [hunk 4]
```diff
@@ -1,3 +1,35 @@
...
+      process.stdout.write(`Bulls: ${result.bulls}, Cows: ${result.cows}\n`);
+
+      if (result.bulls === 4) {
+        process.stdout.write('Congratulations, you got it.');
+        rl.close();
+      } else {
+        prediction();
+      }
+    } else {
+      process.stdout.white('Invalid! Please enter a 4 digits number again.');
```
raulriato, line 29: I believe you misspelled here. It should be `write`

